PR #86, opened 2022-07-13
WIP feat: use firestore instead of env config for kms key storing

capri-xiyue:
use firestore instead of env config for kms key storing


pkg/testutil/firestore_mock.go
```diff
@@ -0,0 +1,285 @@
...
+	// If set, all calls return this error.
+	Error error
+
+	// responses to return if err == nil
+	Resps []proto.Message
+}
+
+func (s *MockFirestoreServer) GetDocument(ctx context.Context, req *firestorepb.GetDocumentRequest) (*firestorepb.Document, error) {
+	md, _ := metadata.FromIncomingContext(ctx)
+	if xg := md["x-goog-api-client"]; len(xg) == 0 || !strings.Contains(xg[0], "gl-go/") {
```
capri-xiyue, line 61: the code borrowed from here https://github.com/googleapis/google-cloud-go/blob/6d1e9fe399d0f9adf1a80107be1b27c725b02571/firestore/apiv1/mock_test.go#L63.
I tried to investigate why we need such metadata validation here, but can't find the clues.

cmd/cert-rotation/main.go
```diff
@@ -88,8 +95,13 @@ func realMain(ctx context.Context) error {
 		return fmt.Errorf("failed to load config: %w", err)
 	}
 
+	fsClient, err := firestore.NewClient(ctx, config.ProjectID)
+	if err != nil {
+		return fmt.Errorf("failed to setup FireSore client: %w", err)
```
raserva, line 100:
```suggestion
		return fmt.Errorf("failed to setup FireStore client: %w", err)
```

cmd/justification/main.go
```diff
@@ -64,12 +65,17 @@ func realMain(ctx context.Context) error {
 		return fmt.Errorf("failed to setup kms client: %w", err)
 	}
 
+	fsClient, err := firestore.NewClient(ctx, cfg.ProjectID)
+	if err != nil {
+		return fmt.Errorf("failed to setup FireSore client: %w", err)
```
raserva, line 70:
```suggestion
		return fmt.Errorf("failed to setup FireStore client: %w", err)
```

cmd/public-key/main.go
```diff
@@ -62,10 +63,15 @@ func realMain(ctx context.Context) error {
 		return fmt.Errorf("failed to load config: %w", err)
 	}
 
+	fsClient, err := firestore.NewClient(ctx, config.ProjectID)
+	if err != nil {
+		return fmt.Errorf("failed to setup FireSore client: %w", err)
```
raserva, line 68:
```suggestion
		return fmt.Errorf("failed to setup FireStore client: %w", err)
```

pkg/config/crypto_config.go
```diff
@@ -45,11 +45,8 @@ type CryptoConfig struct {
 	// DisabledPeriod is a time between when the key is disabled, and when we delete the key.
 	DisabledPeriod time.Duration `yaml:"disabled_period,omitempty" env:"DISABLED_PERIOD,overwrite"`
 
-	// TODO: This is intended to be temporary, and will eventually be retrieved from a persistent external datastore
-	// https://github.com/abcxyz/jvs/issues/17
-	// KeyName format: `projects/*/locations/*/keyRings/*/cryptoKeys/*`
-	// https://pkg.go.dev/google.golang.org/genproto/googleapis/cloud/kms/v1#CryptoKey
-	KeyNames []string `yaml:"key_names,omitempty" env:"KEY_NAMES,overwrite"`
+	// ProjectID is the ID of GCP project where the Firestore documents with the KMS key locates
```
raserva, line 48:
```suggestion
	// ProjectID is the ID of GCP project where the Firestore document with the KMS key is located
```

pkg/testutil/firestore_mock.go
raserva: Seems to be ensuring that the firestore client is identifying itself. Probably unnecessary for us. 

https://cloud.google.com/apis/docs/system-parameters

raserva (commented):
I don't see anything in this change that works with primary key information, is that intended to be done in a future PR?

cmd/cert-rotation/main.go
```diff
@@ -38,12 +40,17 @@ type server struct {
...
+	ctx := r.Context()
+	logger := logging.FromContext(ctx)
 	logger.Info("received request", zap.Any("url", r.URL))
 
 	var errs error
-	// TODO: load keys from DB instead. https://github.com/abcxyz/jvs/issues/17
-	for _, key := range s.handler.CryptoConfig.KeyNames {
+	kmsConfig, err := fsutil.GetKMSConfig(ctx, s.handler.FsClient, fsutil.Collection, fsutil.CertRotationConfigDoc)
+	if err != nil {
+		logger.Error("ran into errors while getting kms config", zap.Error(err))
```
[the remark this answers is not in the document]
capri-xiyue: I was just following the code style of this method https://github.com/abcxyz/jvs/blob/5cc3eb83dc6b7e4af19a40c2c20456d2912770f5/cmd/cert-rotation/main.go#L54. In addition, this `ServeHTTP` method doesn't return error

cmd/cert-rotation/main.go
sethvargo: It's actually "Firestore" 😄

cmd/justification/main.go
```diff
@@ -64,12 +65,17 @@ func realMain(ctx context.Context) error {
 		return fmt.Errorf("failed to setup kms client: %w", err)
 	}
 
+	fsClient, err := firestore.NewClient(ctx, cfg.ProjectID)
```
sethvargo, line 68: I think we need to close these clients when the server shuts down. Applies to KMS too.

pkg/config/crypto_config.go
```diff
@@ -45,11 +45,8 @@ type CryptoConfig struct {
...
+	ProjectID string `yaml:"project_id,omitempty" env:"PROJECT_ID,overwrite"`
```
sethvargo, line 49: This should probably be `FirebaseProjectID`. As-is, we're limiting the ability to have other services in different projects with a generic name.

pkg/config/public_key_config.go
```diff
@@ -19,17 +19,15 @@ import (
...
 
 // PublicKeyConfig is the config used for public key hosting.
 type PublicKeyConfig struct {
-	// TODO: This is intended to be temporary, and will eventually be retrieved from a persistent external datastore
-	// https://github.com/abcxyz/jvs/issues/17
-	// KeyName format: `projects/*/locations/*/keyRings/*/cryptoKeys/*`
-	// https://pkg.go.dev/google.golang.org/genproto/googleapis/cloud/kms/v1#PublicKeyKey
-	KeyNames     []string      `yaml:"key_names,omitempty" env:"KEY_NAMES,overwrite"`
+	// ProjectID is the ID of GCP project where the Firestore documents with the KMS key locates
+	ProjectID    string        `yaml:"project_ID,omitempty" env:"PROJECT_ID,overwrite"`
```
sethvargo, line 30: Same note here: this should be `FirebaseProjectID` or similar.

pkg/firestore/util.go
```diff
@@ -0,0 +1,51 @@
...
+package firestore
+
+import (
+	"context"
+	"fmt"
+
+	"cloud.google.com/go/firestore"
+)
+
+// [START firestore_data_custom_type_definition]
```
sethvargo, line 24: This comment is copied from the code sample - we don't need it.

pkg/firestore/util.go
```diff
@@ -0,0 +1,51 @@
...
+func GetKMSConfig(ctx context.Context, fsClient *firestore.Client, collectionPath, docID string) (*KMSConfig, error) {
+	kSnap, err := fsClient.Collection(collectionPath).Doc(docID).Get(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read from Collection %s Doc %s: %w", collectionPath, docID, err)
+	}
+	var kmsConfig KMSConfig
+	err = kSnap.DataTo(&kmsConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to use firestore document's fields to populate struct: %w", err)
+	}
```
sethvargo, line 49:
```suggestion
	if err := kSnap.DataTo(&kmsConfig); err != nil {
		return nil, fmt.Errorf("failed to use firestore document's fields to populate struct: %w", err)
	}
```

pkg/firestore/util.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2022 Google LLC
```
sethvargo, line 1: Naming things "util" is generally frowned upon in Go.

pkg/justification/processor.go
```diff
@@ -98,7 +102,14 @@ func (p *Processor) CreateToken(ctx context.Context, request *jvspb.CreateJustif
 }
 
 func (p *Processor) getLatestSigner(ctx context.Context) (*signerWithID, error) {
-	ver, err := jvscrypto.GetLatestKeyVersion(ctx, p.kms, p.config.KeyName)
```
sethvargo, line 101: Can we update the comment for `getLatestSigner` to note that it's an expensive operation that communicates with the upstream database?

pkg/firestore/util.go
```diff
@@ -0,0 +1,51 @@
...
+	PublicKeyConfigDoc     = "PublicKeyConfig"
+	CertRotationConfigDoc  = "CertRotationConfig"
+)
+
+type KMSConfig struct {
+	// KeyNames format: `[projects/*/locations/*/keyRings/*/cryptoKeys/*]`
+	KeyNames []string `firestore:"key_names,omitempty"`
+}
+
+// [END firestore_data_custom_type_definition]
```
sethvargo, line 38: Same note, comment is not needed

pkg/firestore/util.go
```diff
@@ -0,0 +1,51 @@
...
+
+func GetKMSConfig(ctx context.Context, fsClient *firestore.Client, collectionPath, docID string) (*KMSConfig, error) {
```
sethvargo, line 40: What is our strategy for getting this initial data into firestore?

sethvargo, line 40: The collection path and document are constants. Why are they also parameters? If the goal is to make them easier to use in tests, I suggest creating a separate function instead for testing.

pkg/jvscrypto/key_hosting.go
```diff
@@ -19,16 +19,19 @@ import (
...
 	"github.com/abcxyz/jvs/pkg/config"
+	fsutil "github.com/abcxyz/jvs/pkg/firestore"
 	"github.com/abcxyz/pkg/cache"
 	"github.com/abcxyz/pkg/logging"
 )
 
 // KeyServer provides all valid and active public keys in a JWKS format.
 type KeyServer struct {
 	KMSClient       *kms.KeyManagementClient
+	FsClient        *firestore.Client
```
sethvargo, line 34: This should be `FirestoreClient`

sethvargo (commented):
Open question: should we default the value of project ID to looking it up in the environment?

pkg/config/public_key_config.go
yolocs: +1

pkg/justification/processor.go
```diff
@@ -40,6 +42,7 @@ import (
 type Processor struct {
 	jvspb.UnimplementedJVSServiceServer
 	kms         *kms.KeyManagementClient
+	fsClient    *firestore.Client
```
yolocs, line 45: We should abstract the implementation away for better encapsulation. So instead of passing in a firestore client, I suggest passing in a `RemoteConfig` (to be defined) interface. The actual implementation will be firestore based.

pkg/firestore/util.go
```diff
@@ -0,0 +1,51 @@
...
+// [START firestore_data_custom_type_definition]
+
+const (
+	Collection             = "JVS"
+	JustificationConfigDoc = "JustificationConfig"
+	PublicKeyConfigDoc     = "PublicKeyConfig"
+	CertRotationConfigDoc  = "CertRotationConfig"
+)
+
+type KMSConfig struct {
```
yolocs, line 33: I think this type is orthogonal to firestore. Thus we should define it in `pkg/config` package.

pkg/firestore/util.go
yolocs, line 40: As I mentioned in the other comment, I suggest refactor this to be `RemoteConfig` interface and firestore implementation.

pkg/justification/processor_test.go
```diff
@@ -105,6 +110,14 @@ func TestCreateToken(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	mockFSClient, mockFSServer, mockFsCleanupFunc, err := testutil.NewMockFS(fakeProjectID)
```
yolocs, line 113: If we do a good job with encapsulating the `RemoteConfig`, then it seems we wouldn't need implement the whole fake firestore.

pkg/jvscrypto/key_hosting.go
yolocs: +1. `FsClient` makes me think of filesystem related stuff

yolocs (changes requested):
In the future, try also break up the PR. E.g. the firestore config client, the fake could be its own PR.

capri-xiyue:
> I don't see anything in this change that works with primary key information, is that intended to be done in a future PR?

Do you. mean for [SetPrimary](https://github.com/abcxyz/jvs/blob/5cc3eb83dc6b7e4af19a40c2c20456d2912770f5/pkg/jvscrypto/kmsutil.go#L198) and [getPrimary](https://github.com/abcxyz/jvs/blob/5cc3eb83dc6b7e4af19a40c2c20456d2912770f5/pkg/jvscrypto/kmsutil.go#L184), we want to store the primary key info in the firestore instead of relying on the KMS key labels, right?

raserva:
> > I don't see anything in this change that works with primary key information, is that intended to be done in a future PR?
> 
> Do you mean for [SetPrimary](https://github.com/abcxyz/jvs/blob/5cc3eb83dc6b7e4af19a40c2c20456d2912770f5/pkg/jvscrypto/kmsutil.go#L198) and [getPrimary](https://github.com/abcxyz/jvs/blob/5cc3eb83dc6b7e4af19a40c2c20456d2912770f5/pkg/jvscrypto/kmsutil.go#L184), we want to store the primary key info in the firestore instead of relying on the KMS key labels, right?

Correct. Right now we're storing primary information on key labels, but it should instead be stored in firebase.

capri-xiyue:
> > > I don't see anything in this change that works with primary key information, is that intended to be done in a future PR?
> > 
> > 
> > Do you mean for [SetPrimary](https://github.com/abcxyz/jvs/blob/5cc3eb83dc6b7e4af19a40c2c20456d2912770f5/pkg/jvscrypto/kmsutil.go#L198) and [getPrimary](https://github.com/abcxyz/jvs/blob/5cc3eb83dc6b7e4af19a40c2c20456d2912770f5/pkg/jvscrypto/kmsutil.go#L184), we want to store the primary key info in the firestore instead of relying on the KMS key labels, right?
> 
> Correct. Right now we're storing primary information on key labels, but it should instead be stored in firebase. That can be done as a follow-up if you want.
> 
> But one thing that will be important to think about even in this PR is that we cannot load the config once and be done, instead we need to make sure we have any updates from firebase when we are signing tokens to ensure that we're using the most recent primary

Looks like firestore supports listen to multiple documents(https://firebase.google.com/docs/firestore/query-data/listen#listen_to_multiple_documents_in_a_collection) But in our case, we need long-running listen. Looks like we may have to write some logic to achieve it.

yolocs:
Found https://github.com/spf13/viper#firestore. Can we use that? We already use viper for CLI config.

capri-xiyue:
closed because I created a new PR https://github.com/abcxyz/jvs/pull/91

PR #87, opened 2022-07-14
feat: add additional metrics to the dashboard

yolocs:
- Added CPU/Mem usage and container instance count metrics
- Uniformly used LINE plotting and fixed some align types

image

### DIFF
--- a/terraform/modules/monitoring/dashboard.json
+++ b/terraform/modules/monitoring/dashboard.json
@@ -15,7 +15,7 @@
             "dataSets": [
               {
                 "minAlignmentPeriod": "60s",
-                "plotType": "STACKED_BAR",
+                "plotType": "LINE",
                 "targetAxis": "Y1",
                 "timeSeriesQuery": {
                   "apiSource": "DEFAULT_CLOUD",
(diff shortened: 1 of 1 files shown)


terraform/modules/monitoring/dashboard.json
```diff
@@ -173,8 +277,122 @@
...
+                "minAlignmentPeriod": "60s",
+                "plotType": "LINE",
+                "targetAxis": "Y1",
+                "timeSeriesQuery": {
+                  "apiSource": "DEFAULT_CLOUD",
+                  "timeSeriesFilter": {
+                    "aggregation": {
+                      "alignmentPeriod": "60s",
+                      "crossSeriesReducer": "REDUCE_PERCENTILE_99",
+                      "perSeriesAligner": "ALIGN_SUM"
```
raserva, line 302: do we want to sum up the memory usages? This should probably be 99th percentage or avg

raserva (commented):
one thing that i notice is that it is a bit hard to differentiate between JVS and certifcate metrics. Is is possible to break them up into separate sections? Or have some other mechanism for logically separating them

terraform/modules/monitoring/dashboard.json
raserva: I do see the reducer is set to 99th. its not clear to me the difference between reducer and aligner

yolocs: Me neither. I simply used the same parameters that cloud run metrics tab has. I suppose it's the good version.
I also didn't find a way to logically separate metrics. Need more learning. Should we get this in first?

raserva: Sure, for now we can merge as-is.

raserva (approved):

PR #88, opened 2022-07-14
Update envconfig defaults processing

sethvargo:
This updates to the latest version of envconfig which correctly handles not overwriting default values on input structs. This simplifies a lot of the SetDefault() logic that existed to get around this limitation.

As part of this change, I noticed there was inconsistent enforcement of the "Version" field in structs, and that the field was limited to integer values (which would prefer us from having a "v1" or "1.0-beta" version in the future). I created a small type that lets us define and manage allowed versions, and standardized on the error messages across packages. I also updated many of the error messages to be more specific about the failure.

Finally, I removed the envconfig annotations from CLIConfig, since they were actually unused; Viper handles the environment variables and initialization for this structure. For the other configs, I added dedicated tests for defaults.


pkg/config/cli_config.go
```diff
@@ -20,17 +20,15 @@ import (
...
-const (
-	// DefaultCLIConfigVersion is the default CLI config version.
-	DefaultCLIConfigVersion = 1
-)
+// CLIConfigVersions is the list of allowed versions for the CLIConfig.
+var CLIConfigVersions = NewVersionList("1")
 
 type CLIConfig struct {
 	// Version is the version of the config.
-	Version uint8 `yaml:"version,omitempty" env:"VERSION,overwrite"`
```
sethvargo, line 30: We don't actually use envconfig here, we use Viper.

pkg/config/versions.go
```diff
@@ -0,0 +1,61 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+import "sort"
+
+// VersionList is a set of allowed versions. Create with NewVersionList.
+type VersionList struct {
```
raserva, line 20: nit: this seems generic enough that it could probably live in pkg

raserva (approved):

pkg/config/versions.go
sethvargo: Yea we can move it later if we want to adopt it in lumberjack too.

PR #95, opened 2022-07-26
fix: Modify JwkProviderBuilder call

raserva:
if you pass in the string directly, it appends "/.well-known/jwks.json" on the end. This means that if we have the configuration the same as golang, we end up with the path "/.well-known/jwks/.well-known/jwks.json". This changes it to pass in a URL instead, which JwksProvider does not append a path on the end. see https://javadoc.io/doc/com.auth0/jwks-rsa/latest/com/auth0/jwk/JwkProviderBuilder.html

### DIFF
--- a/client-lib/java/src/main/java/com/abcxyz/jvs/JVSClientBuilder.java
+++ b/client-lib/java/src/main/java/com/abcxyz/jvs/JVSClientBuilder.java
@@ -24,6 +24,8 @@ import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import com.google.common.base.Strings;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.time.Duration;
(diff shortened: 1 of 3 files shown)


yolocs (approved):

client-lib/java/src/main/java/com/abcxyz/jvs/JVSClientBuilder.java
```diff
@@ -101,8 +103,18 @@ public JvsClient build() {
...
     configuration.validate();
 
+    URL url;
+    try {
+      // We have to convert endpoint to URL, as the JwkProviderBuilder
+      // appends its own path on the end if you simply pass the string.
+      url = new URL(configuration.getJvsEndpoint());
+    } catch (MalformedURLException e) {
+      throw new IllegalArgumentException(
+          String.format("endpoint was invalid %s", configuration.getJvsEndpoint()));
```
sethvargo, line 113: Should we say something like "endpoint must be of the format http(s)://domain[:port]"?

PR #96, opened 2022-07-26
feat: remove env config in public key api and cert rotation api

capri-xiyue:
remove env config in public key api and cert rotation api

### DIFF
--- a/cmd/cert-rotation/main.go
+++ b/cmd/cert-rotation/main.go
@@ -24,8 +24,11 @@ import (
 	"syscall"
 	"time"
 
+	"cloud.google.com/go/firestore"
+
+	"github.com/abcxyz/jvs/pkg/util"
+
 	kms "cloud.google.com/go/kms/apiv1"
-	"github.com/abcxyz/jvs/pkg/cleanup"
 	"github.com/abcxyz/jvs/pkg/config"
 	"github.com/abcxyz/jvs/pkg/jvscrypto"
 	"github.com/abcxyz/pkg/logging"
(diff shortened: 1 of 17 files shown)


cmd/cert-rotation/main.go
```diff
@@ -82,16 +90,29 @@ func realMain(ctx context.Context) error {
...
-	defer cleanup.GracefulClose(logger, kmsClient)
+	defer util.GracefulClose(logger, kmsClient)
 
-	config, err := config.LoadCryptoConfig(ctx, []byte{})
+	cryptoCfg, err := config.LoadCryptoConfig(ctx, []byte{})
 	if err != nil {
 		return fmt.Errorf("failed to load config: %w", err)
 	}
 
+	firestoreDoc, err := util.ParseFirestoreDocResource(cryptoCfg.FirestoreDocResourceName)
```
sethvargo, line 100: I don't know if we want this to be firestore per the doc just shared out. I'm worried about the coupling here.

yolocs: If we're unsure about firestore, maybe we should defer using firestore at all. It's not like we're missing functionality without it. Thoughts?

sethvargo: Agree on deferring

capri-xiyue: Ok. I will revert the previous PR which already uses firestore https://github.com/abcxyz/jvs/pull/91

capri-xiyue:
closed because we are deferring the firestore

PR #100, opened 2022-07-27
feat: add support for break-glass tokens in java

raserva:


client-lib/java/src/main/java/com/abcxyz/jvs/JvsConfiguration.java
```diff
@@ -40,6 +40,9 @@ public class JvsConfiguration {
   @JsonProperty("cache_timeout")
   private Duration cacheTimeout = Duration.ofMinutes(5);
 
+  @JsonProperty("allow_breakglass")
+  private boolean breakglassAllowed = false;
+
```
yolocs, line 45: Should the default be `true`? Or change it to `breakglassDisallowed` and default to `false`. I think breakglass should be supported by default.

raserva: Sure, i can change this if we think opt-out is better than opt-in for breakglass

yolocs (approved):

myurtoglu (approved):

PR #107, opened 2022-08-03
fix: use primary key version instead of latest key version to sign the token

### LINKED ISSUE #106: Sign token with primary instead of latest version
### TL;DR

Sign token with primary instead of latest version
https://github.com/abcxyz/jvs/blob/23cdda185a8fc277ba798bfaf3e34543458f1934/pkg/justification/processor.go#L100

### Expected behavior

_No response_

### Observed behavior

_No response_

### Environment Details

```markdown
None
```

### Additional information

_No response_

capri-xiyue:
fix #106 
1. use primary key version instead of latest key version to sign the token 
2. fix main brnach lint error(ran into multiple lint errors from main branch)

### DIFF
--- a/pkg/justification/processor_test.go
+++ b/pkg/justification/processor_test.go
@@ -158,7 +158,7 @@ func TestCreateToken(t *testing.T) {
 			}
 
 			mockKeyManagement.VersionName = version
-			mockKeyManagement.Labels["primary"] = "ver_" + "[VERSION]" + "-0"
+			mockKeyManagement.Labels[jvscrypto.PrimaryKey] = jvscrypto.PrimaryLabelPrefix + "[VERSION]" + "-0"
 
 			authKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
 			if err != nil {
--- a/pkg/justification/processor_test.go
+++ b/pkg/justification/processor_test.go
@@ -60,9 +60,6 @@ func TestCreateToken(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	key := "projects/[PROJECT]/locations/[LOCATION]/keyRings/[KEY_RING]/cryptoKeys/[CRYPTO_KEY]"
-	version := key + "/cryptoKeyVersions/[VERSION]"
-	keyID := key + "/cryptoKeyVersions/[VERSION]-0"
 
 	tests := []struct {
 		name      string
@@ -108,6 +105,9 @@ func TestCreateToken(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
 			var clientOpt option.ClientOption
+			key := "projects/[PROJECT]/locations/[LOCATION]/keyRings/[KEY_RING]/cryptoKeys/[CRYPTO_KEY]"
+			version := key + "/cryptoKeyVersions/[VERSION]"
+			keyID := key + "/cryptoKeyVersions/[VERSION]-0"
 			mockKeyManagement := &testutil.MockKeyManagementServer{
 				UnimplementedKeyManagementServiceServer: kmspb.UnimplementedKeyManagementServiceServer{},
 				Reqs:                                    make([]proto.Message, 1),
--- a/pkg/justification/processor_test.go
+++ b/pkg/justification/processor_test.go
@@ -56,98 +56,10 @@ func TestCreateToken(t *testing.T) {
 	t.Parallel()
 	ctx := context.Background()
 
-	var clientOpt option.ClientOption
-	mockKeyManagement := &testutil.MockKeyManagementServer{
-		UnimplementedKeyManagementServiceServer: kmspb.UnimplementedKeyManagementServiceServer{},
-		Reqs:                                    make([]proto.Message, 1),
-		Err:                                     nil,
-		Resps:                                   make([]proto.Message, 1),
-		NumVersions:                             1,
-		Labels:                                  make(map[string]string),
-	}
-
-	privateKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
-	if err != nil {
-		t.Fatal(err)
-	}
-	mockKeyManagement.PrivateKey = privateKey
-	x509EncodedPub, err := x509.MarshalPKIXPublicKey(privateKey.Public())
-	if err != nil {
-		t.Fatal(err)
-	}
-	pemEncodedPub := pem.EncodeToMemory(&pem.Block{Type: "PUBLIC KEY", Bytes: x509EncodedPub})
-	mockKeyManagement.PublicKey = string(pemEncodedPub)
-
-	serv := grpc.NewServer()
-	kmspb.RegisterKeyManagementServiceServer(serv, mockKeyManagement)
-
-	lis, err := net.Listen("tcp", "localhost:0")
-	if err != nil {
-		t.Fatal(err)
-	}
-	// not checked, but makes linter happy
-	errs := make(chan error, 1)
-	go func() {
-		errs <- serv.Serve(lis)
-		close(errs)
-	}()
-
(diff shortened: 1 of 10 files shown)


test/integ/main_test.go
```diff
@@ -244,6 +245,7 @@ func TestJVS(t *testing.T) {
 	}
 }
 
+//nolint:tparallel // subtests need to run in sequence.
```
sethvargo, line 248: Can you elaborate on why?

test/integ/main_test.go
```diff
@@ -736,12 +742,14 @@ func testIsIntegration(tb testing.TB) bool {
...
 // Create an asymmetric signing key for use with integration tests.
-func testCreateKey(ctx context.Context, tb testing.TB, kmsClient *kms.KeyManagementClient, keyRing string) string {
+func testCreateKey(ctx context.Context, tb testing.TB, kmsClient *kms.KeyManagementClient, keyRing, primaryKeyVersion string) string {
 	tb.Helper()
 	u, err := uuid.NewUUID()
 	if err != nil {
 		tb.Fatalf("failed to create uuid : %s", err)
 	}
+	labels := make(map[string]string)
+	labels["primary"] = "ver_" + primaryKeyVersion
```
sethvargo, line 752: 1. It would be better to just make the map: `labels := map[string]string{"primary": "ver_"+primaryKeyVersion}`

2. Why do we need to store the primary version on a label? There's already a `primary` field on the proto which links to the specific version: https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings.cryptoKeys#CryptoKey

test/integ/main_test.go
capri-xiyue: It was written in https://github.com/abcxyz/jvs/blob/23cdda185a8fc277ba798bfaf3e34543458f1934/test/integ/main_test.go#L280
Should I move the original comment to here to make it clear?

test/integ/main_test.go
capri-xiyue: Why do we need to store the primary version on a label? 
JVS use labels to store the primary version https://github.com/abcxyz/jvs/blob/23cdda185a8fc277ba798bfaf3e34543458f1934/pkg/jvscrypto/kmsutil.go#L198 instead of the `primary` field. Do we want to change it to the `primary` field now?

test/integ/main_test.go
sethvargo: Yes, please

test/integ/main_test.go
yolocs: We should definitely use the `primary` field in the KMS API. I'm surprised we didn't discover that in the first place.

sethvargo: @raserva do you remember why we used labels to store the primary version instead of the `primary` proto field?

It would be better to use the `primary` field directly, since fetching the `cryptoKey` will then return the embedded `cyrptoKeyVersion` directly in the same API call. There's a separate API method for setting the primary version: https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings.cryptoKeys/updatePrimaryVersion

raserva: primary isn't supported for async keys.

raserva: https://cloud.google.com/kms/docs/key-states#symmetric_encryption

raserva: Trying to find where that is explicitly stated, but i did try to use that field.

raserva: This is called out in [the original PR](https://github.com/abcxyz/jvs/pull/25). 

> Unfortunately, there is no standard way in KMS to mark an asymmetric key as primary (there is a primary field, but it is unsupported for asymettric keys). Therefore, we need to store this information somewhere else. This solves this issue by utilizing [KMS key labels](https://cloud.google.com/kms/docs/creating-managing-labels) primary, new, old. This allows us to distinguish between new key versions and handle odd cases gracefully, such as rollbacks.

raserva: Here you go. 
```
rsrv@rsrv-desktop:~$ gcloud kms keys set-primary-version <mykey> --version=1 --location=global --keyring=ci-keyring
ERROR: (gcloud.kms.keys.set-primary-version) FAILED_PRECONDITION: Keys with purpose ASYMMETRIC_SIGN do not have a primary version.
- '@type': type.googleapis.com/google.rpc.PreconditionFailure
  violations:
  - subject: <mykey>
    type: CRYPTO_SCHEME_MISMATCH
- '@type': type.googleapis.com/google.rpc.DebugInfo
  detail: '[ORIGINAL ERROR] generic::failed_precondition: Keys with purpose ASYMMETRIC_SIGN
    do not have a primary version. [google.rpc.error_details_ext] { message: "Keys
    with purpose ASYMMETRIC_SIGN do not have a primary version." details { [type.googleapis.com/google.rpc.PreconditionFailure]
    { violations { type: "CRYPTO_SCHEME_MISMATCH" subject: "<mykey>"
    } } } }'
```

capri-xiyue: I found docs here: https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings.cryptoKeys
`Keys with [purpose](https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings.cryptoKeys#CryptoKey.FIELDS.purpose) [ENCRYPT_DECRYPT](https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings.cryptoKeys#CryptoKeyPurpose.ENUM_VALUES.ENCRYPT_DECRYPT) may have a primary. For other keys, this field will be omitted.`

capri-xiyue: I will add a comment in setting primary method to explicitly list the reason why we don't use the `primary` field.

sethvargo: Let's document that rationale inline 😄

test/integ/main_test.go
capri-xiyue: done

pkg/jvscrypto/kmsutil.go
```diff
@@ -180,21 +180,23 @@ func SignToken(token *jwt.Token, signer crypto.Signer) (string, error) {
...
+	if primary, ok := response.Labels[PrimaryKey]; ok {
+		primary = strings.TrimPrefix(primary, PrimaryLabelPrefix)
 		return fmt.Sprintf("%s/cryptoKeyVersions/%s", key, primary), nil
 	}
 	// no primary found
 	return "", nil
 }
 
 // SetPrimary sets the key version name as primary in the key labels.
+// 'Primary' field will be omitted for keys with purpose other than ENCRYPT_DECRYPT(https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings.cryptoKeys).
```
yolocs, line 198: nit: also mention what's the key type we're using.

pkg/justification/processor_test.go
```diff
@@ -188,6 +103,95 @@ func TestCreateToken(t *testing.T) {
 	for _, tc := range tests {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			var clientOpt option.ClientOption
+			key := "projects/[PROJECT]/locations/[LOCATION]/keyRings/[KEY_RING]/cryptoKeys/[CRYPTO_KEY]"
+			version := key + "/cryptoKeyVersions/[VERSION]"
+			keyID := key + "/cryptoKeyVersions/[VERSION]-0"
+			mockKeyManagement := &testutil.MockKeyManagementServer{
```
yolocs, line 111: Great we can make it parallel!

Could you do a refactor to make use of https://github.com/abcxyz/pkg/blob/main/testutil/grpc.go?

yolocs (commented):
All other looks good to me

pkg/jvscrypto/kmsutil.go
capri-xiyue: DONE

yolocs (approved):

PR #110, opened 2022-08-04
feat: added breakglass support in go client

### LINKED ISSUE #109: support breakglass in go client
### TL;DR

support breakglass in go client

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

capri-xiyue:
fix #109 
1. added breakglass support in go client
2. fixed lint error in main branch.

### DIFF
--- a/client-lib/go/client/jvs_client.go
+++ b/client-lib/go/client/jvs_client.go
@@ -80,7 +80,7 @@ func (j *JVSClient) ValidateJWT(jwtStr string) (*jwt.Token, error) {
 }
 
 func (j *JVSClient) unsignedTokenValidAndAllowed(token jwt.Token) error {
-	if j.config.ForbidBreakglass {
+	if !j.config.AllowBreakglass {
 		return fmt.Errorf("breakglass is forbidden, denying")
 	}
 
--- a/client-lib/go/client/jvs_client.go
+++ b/client-lib/go/client/jvs_client.go
@@ -97,7 +97,7 @@ func (j *JVSClient) unsignedTokenValidAndAllowed(token jwt.Token) error {
 	if err = json.Unmarshal(justsBytes, &justifications); err != nil {
 		return fmt.Errorf("failed to unmarshal 'justs', denying")
 	}
-	
+
 	for _, justification := range justifications {
 		if justification.GetCategory() == BreakglassCategory {
 			return nil
--- a/client-lib/go/client/jvs_client.go
+++ b/client-lib/go/client/jvs_client.go
@@ -71,8 +71,7 @@ func (j *JVSClient) ValidateJWT(jwtStr string) (*jwt.Token, error) {
 		if err != nil {
 			return nil, fmt.Errorf("failed to parse jwt %s: %w", jwtStr, err)
 		}
-		err = j.unsignedTokenValidAndAllowed(token)
-		if err != nil {
+		if err := j.unsignedTokenValidAndAllowed(token); err != nil {
 			return nil, fmt.Errorf("token unsigned and could not be validated: %w", err)
 		}
 		return &token, nil
@@ -94,7 +93,7 @@ func (j *JVSClient) unsignedTokenValidAndAllowed(token jwt.Token) error {
 		return fmt.Errorf("failed to marshal 'justs', denying")
 	}
 	var justifications []*jvsapis.Justification
-	if err = json.Unmarshal(justsBytes, &justifications); err != nil {
+	if err := json.Unmarshal(justsBytes, &justifications); err != nil {
 		return fmt.Errorf("failed to unmarshal 'justs', denying")
 	}
 
--- a/client-lib/go/client/jvs_client.go
+++ b/client-lib/go/client/jvs_client.go
@@ -23,7 +23,6 @@ import (
 
 	jvspb "github.com/abcxyz/jvs/apis/v0"
 	"github.com/abcxyz/jvs/pkg/jvscrypto"
-
 	"github.com/lestrrat-go/jwx/v2/jwk"
 	"github.com/lestrrat-go/jwx/v2/jwt"
 )
@@ -68,19 +67,17 @@ func NewJVSClient(ctx context.Context, config *JVSConfig) (*JVSClient, error) {
 
 // ValidateJWT takes a jwt string, converts it to a JWT, and validates the signature.
 func (j *JVSClient) ValidateJWT(jwtStr string) (*jwt.Token, error) {
-	token, err := jwt.Parse([]byte(jwtStr), jwt.WithVerify(false))
-	if err != nil {
-		return nil, fmt.Errorf("failed to verify jwt %s: %w", jwtStr, err)
-	}
-
 	// Handle Break-glass tokens.
 	if strings.HasSuffix(jwtStr, UnsignedPostfix) {
+		token, err := jwt.Parse([]byte(jwtStr), jwt.WithVerify(false))
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse jwt %s: %w", jwtStr, err)
+		}
 		valid, err := j.unsignedTokenValidAndAllowed(token)
-		if valid {
-			return &token, nil
-		} else {
+		if !valid {
 			return nil, fmt.Errorf("token unsigned and could not be validated: %w", err)
 		}
+		return &token, nil
 	}
 	return jvscrypto.ValidateJWT(j.keys, jwtStr)
 }
--- a/client-lib/go/client/jvs_client.go
+++ b/client-lib/go/client/jvs_client.go
@@ -71,8 +71,8 @@ func (j *JVSClient) ValidateJWT(jwtStr string) (*jwt.Token, error) {
 		if err != nil {
 			return nil, fmt.Errorf("failed to parse jwt %s: %w", jwtStr, err)
 		}
-		valid, err := j.unsignedTokenValidAndAllowed(token)
-		if !valid {
+		err = j.unsignedTokenValidAndAllowed(token)
+		if err != nil {
 			return nil, fmt.Errorf("token unsigned and could not be validated: %w", err)
 		}
 		return &token, nil
@@ -80,27 +80,27 @@ func (j *JVSClient) ValidateJWT(jwtStr string) (*jwt.Token, error) {
 	return jvscrypto.ValidateJWT(j.keys, jwtStr)
 }
 
-func (j *JVSClient) unsignedTokenValidAndAllowed(token jwt.Token) (bool, error) {
+func (j *JVSClient) unsignedTokenValidAndAllowed(token jwt.Token) error {
 	if j.config.ForbidBreakglass {
-		return false, fmt.Errorf("breakglass is forbidden, denying")
+		return fmt.Errorf("breakglass is forbidden, denying")
 	}
 	justs, ok := token.Get("justs")
 	if !ok {
-		return false, fmt.Errorf("can't find 'justs' in claims, denying")
+		return fmt.Errorf("can't find 'justs' in claims, denying")
 	}
 	justsBytes, err := json.Marshal(justs)
 	if err != nil {
-		return false, fmt.Errorf("failed to marshal 'justs', denying")
+		return fmt.Errorf("failed to marshal 'justs', denying")
 	}
 	var justifications []*jvsapis.Justification
 	err = json.Unmarshal(justsBytes, &justifications)
 	if err != nil {
-		return false, fmt.Errorf("failed to unmarshal 'justs', denying")
+		return fmt.Errorf("failed to unmarshal 'justs', denying")
 	}
 	for _, justification := range justifications {
 		if justification.GetCategory() == BreakglassCategory {
-			return true, nil
+			return nil
 		}
 	}
-	return false, fmt.Errorf("justification category is not breakglass, denying")
+	return fmt.Errorf("justification category is not breakglass, denying")
 }
--- a/client-lib/go/client/jvs_client.go
+++ b/client-lib/go/client/jvs_client.go
@@ -84,6 +84,7 @@ func (j *JVSClient) unsignedTokenValidAndAllowed(token jwt.Token) error {
 	if j.config.ForbidBreakglass {
 		return fmt.Errorf("breakglass is forbidden, denying")
 	}
+
 	justs, ok := token.Get("justs")
 	if !ok {
 		return fmt.Errorf("can't find 'justs' in claims, denying")
@@ -93,10 +94,10 @@ func (j *JVSClient) unsignedTokenValidAndAllowed(token jwt.Token) error {
 		return fmt.Errorf("failed to marshal 'justs', denying")
 	}
 	var justifications []*jvsapis.Justification
-	err = json.Unmarshal(justsBytes, &justifications)
-	if err != nil {
+	if err = json.Unmarshal(justsBytes, &justifications); err != nil {
 		return fmt.Errorf("failed to unmarshal 'justs', denying")
 	}
+	
 	for _, justification := range justifications {
 		if justification.GetCategory() == BreakglassCategory {
 			return nil
--- a/client-lib/go/client/jvs_client.go
+++ b/client-lib/go/client/jvs_client.go
@@ -17,18 +17,28 @@ package client
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"strings"
 
+	jvspb "github.com/abcxyz/jvs/apis/v0"
 	"github.com/abcxyz/jvs/pkg/jvscrypto"
+
 	"github.com/lestrrat-go/jwx/v2/jwk"
 	"github.com/lestrrat-go/jwx/v2/jwt"
+)
+
+const (
+	UnsignedPostfix    = ".NOT_SIGNED"
+	BreakglassCategory = "breakglass"
 )
 
 // JVSClient allows for getting JWK keys from the JVS and validating JWTs with
 // those keys.
 type JVSClient struct {
-	config *JVSConfig
-	keys   jwk.Set
+	config           *JVSConfig
+	keys             jwk.Set
+	forbidBreakglass bool
 }
 
 // NewJVSClient returns a JVSClient with the cache initialized.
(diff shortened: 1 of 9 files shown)


test/integ/main_test.go
```diff
@@ -338,8 +339,7 @@ func TestRotator(t *testing.T) {
 	})
 }
 
-//nolint:tparallel
-// Subtests need to run in sequence. To parallelize this, but we'd need separate keys from the above (more cruft).
+//nolint:tparallel // Subtests need to run in sequence. To parallelize this, but we'd need separate keys from the above (more cruft).
```
raserva, line 342: nit: 2 comments combined

raserva (approved):

client-lib/go/client/jvs_client.go
```diff
@@ -17,18 +17,27 @@ package client
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"strings"
 
+	jvspb "github.com/abcxyz/jvs/apis/v0"
```
yolocs, line 24: nit: use `jvsapi` as the alias

client-lib/go/client/jvs_client.go
```diff
@@ -50,12 +59,50 @@ func NewJVSClient(ctx context.Context, config *JVSConfig) (*JVSClient, error) {
 	}
 
 	return &JVSClient{
-		config: config,
-		keys:   cached,
+		config:           config,
+		keys:             cached,
+		forbidBreakglass: config.ForbidBreakglass,
```
yolocs, line 64: You already have `config` as a field, why extracting this as a separate field?

client-lib/go/client/jvs_client.go
```diff
@@ -50,12 +59,50 @@ func NewJVSClient(ctx context.Context, config *JVSConfig) (*JVSClient, error) {
...
+		if !valid {
+			return nil, fmt.Errorf("token unsigned and could not be validated: %w", err)
+		}
+		return &token, nil
+	}
 	return jvscrypto.ValidateJWT(j.keys, jwtStr)
 }
+
+func (j *JVSClient) unsignedTokenValidAndAllowed(token jwt.Token) (bool, error) {
+	if j.forbidBreakglass {
```
yolocs, line 86: You can do the following instead: 

```suggestion
	if j.config.ForbidBreakglass {
```

client-lib/go/client/jvs_client.go
```diff
@@ -50,12 +59,50 @@ func NewJVSClient(ctx context.Context, config *JVSConfig) (*JVSClient, error) {
...
+		return false, fmt.Errorf("break glass tokens is forbidden, denying")
```
yolocs, line 87:
```suggestion
		return false, fmt.Errorf("breakglass is forbidden, denying")
```

client-lib/go/client/jvs_client.go
```diff
@@ -50,12 +59,50 @@ func NewJVSClient(ctx context.Context, config *JVSConfig) (*JVSClient, error) {
...
+	}
+	justsBytes, err := json.Marshal(justs)
+	if err != nil {
+		return false, fmt.Errorf("failed to marshal 'justs', denying")
+	}
+	var justifications []*jvspb.Justification
+	err = json.Unmarshal(justsBytes, &justifications)
+	if err != nil {
+		return false, fmt.Errorf("failed to unmarshal 'justs', denying")
+	}
```
yolocs, line 101:
```suggestion
	if err := json.Unmarshal(justsBytes, &justifications); err != nil {
		return false, fmt.Errorf("failed to unmarshal 'justs', denying")
	}
```

client-lib/go/client/jvs_client.go
```diff
@@ -50,12 +59,50 @@ func NewJVSClient(ctx context.Context, config *JVSConfig) (*JVSClient, error) {
...
+	for _, justification := range justifications {
+		if justification.GetCategory() == BreakglassCategory {
+			return true, nil
+		}
+	}
+	return false, fmt.Errorf("unable to find correct break-glass category, denying")
```
yolocs, line 107:
```suggestion
	return false, fmt.Errorf("justification category is not breakglass, denying")
```

client-lib/go/client/jvs_client.go
```diff
@@ -50,12 +59,50 @@ func NewJVSClient(ctx context.Context, config *JVSConfig) (*JVSClient, error) {
...
+		return false, fmt.Errorf("failed to marshal 'justs', denying")
+	}
+	var justifications []*jvspb.Justification
+	err = json.Unmarshal(justsBytes, &justifications)
+	if err != nil {
+		return false, fmt.Errorf("failed to unmarshal 'justs', denying")
+	}
+	for _, justification := range justifications {
+		if justification.GetCategory() == BreakglassCategory {
+			return true, nil
```
yolocs, line 104: Should we expect all justifications to have `breakglass` category?

client-lib/go/client/jvs_client_test.go
```diff
@@ -109,32 +86,54 @@ func TestValidateJWT(t *testing.T) {
...
 			jwt:       validJWT2,
 			wantToken: tok2,
-		}, {
+		},
+		{
 			name:    "unsigned",
 			jwt:     unsignedJWT,
-			wantErr: "required field \"signatures\" not present",
-		}, {
+			wantErr: "unable to find correct break-glass category, denying",
```
yolocs, line 123:
```suggestion
			wantErr: "unable to find correct breakglass category, denying",
```

client-lib/go/client/jvs_client_test.go
```diff
@@ -144,6 +143,27 @@ func TestValidateJWT(t *testing.T) {
...
+			mux.HandleFunc(path, func(w http.ResponseWriter, r *http.Request) {
+				w.WriteHeader(http.StatusOK)
+				fmt.Fprintf(w, "%s", j)
+			})
+
+			svr := httptest.NewServer(mux)
+
+			t.Cleanup(func() {
+				svr.Close()
+			})
```
yolocs, line 157: What's the reason to move these lines into the subtest?

client-lib/go/client/jvs_client_test.go
```diff
@@ -109,32 +86,54 @@ func TestValidateJWT(t *testing.T) {
...
 
-	invalidSignatureJWT := unsignedJWT + sig // signature from a different JWT
+	invalidSignatureJWT := string(unsig) + sig // signature from a different JWT
+
+	breakglassTok := createBreakglassToken(t, "test_id_3")
+	breakglassTokStr, err := jwt.NewSerializer().Serialize(breakglassTok)
+	if err != nil {
+		t.Fatal("Couldn't get breakglass token string.")
+	}
+	breakglassJWT := string(breakglassTokStr) + ".NOT_SIGNED"
```
yolocs, line 101: What's the reason to _not_ wrap all lines here in the `createBreakglassToken` func?

yolocs (changes requested):

test/integ/main_test.go
capri-xiyue: This is the lint format: https://golangci-lint.run/usage/false-positives/

client-lib/go/client/jvs_client_test.go
capri-xiyue: The jvs client need to use tb.forbidBreakglass. Therefore, I move the initialization of jvs client into the subtest.

test/integ/main_test.go
capri-xiyue: //nolint:gocyclo // This legacy function is complex but the team too busy to simplify it
func someLegacyFunction() *string {
  // ...
}

client-lib/go/client/jvs_client.go
capri-xiyue: Thank you for catching it! We don't need it. At first, I just followed the java code logic and extracted the boolean as a separate filed https://github.com/abcxyz/jvs/blob/main/client-lib/java/src/main/java/com/abcxyz/jvs/JvsClient.java#L44

client-lib/go/client/jvs_client.go
capri-xiyue: done. I also updated the message in java client to make it consistent.

client-lib/go/client/jvs_client.go
capri-xiyue: The logic in java client only check one justification to have a `breakglass` category.
https://github.com/abcxyz/jvs/blob/3b5e51097c09a19a4f2ecfff5be9c6ccb1fcdc3e/client-lib/java/src/main/java/com/abcxyz/jvs/JvsClient.java#L78

client-lib/go/client/jvs_client_test.go
```diff
@@ -109,32 +86,54 @@ func TestValidateJWT(t *testing.T) {
...
 	split := strings.Split(validJWT2, ".")
 	sig := split[len(split)-1]
 
-	invalidSignatureJWT := unsignedJWT + sig // signature from a different JWT
+	invalidSignatureJWT := string(unsig) + sig // signature from a different JWT
+
+	breakglassTok := createBreakglassToken(t, "test_id_3")
+	breakglassTokStr, err := jwt.NewSerializer().Serialize(breakglassTok)
+	if err != nil {
+		t.Fatal("Couldn't get breakglass token string.")
```
sethvargo, line 99: Log the actual error returned

client-lib/go/client/jvs_config.go
```diff
@@ -37,6 +37,9 @@ type JVSConfig struct {
 
 	// CacheTimeout is the duration that keys stay in cache before being revoked.
 	CacheTimeout time.Duration `yaml:"cache_timeout" env:"CACHE_TIMEOUT,overwrite,default=5m"`
+
+	// ForbidBreakglass represents whether the jvs client forbids breakglass.
+	ForbidBreakglass bool `yaml:"forbid_breakglass" env:"FORBID_BREAKGLASS,overwrite,default=false"`
```
sethvargo, line 42: I think the default should be to forbid breakglass. Users should have to opt-in to allowing breakglass. As such, we should rename this to `AllowBreakglass`, but keep the default as false. @yolocs

client-lib/go/client/jvs_client.go
```diff
@@ -50,12 +59,50 @@ func NewJVSClient(ctx context.Context, config *JVSConfig) (*JVSClient, error) {
...
+		valid, err := j.unsignedTokenValidAndAllowed(token)
+		if !valid {
+			return nil, fmt.Errorf("token unsigned and could not be validated: %w", err)
+		}
+		return &token, nil
+	}
 	return jvscrypto.ValidateJWT(j.keys, jwtStr)
 }
+
+func (j *JVSClient) unsignedTokenValidAndAllowed(token jwt.Token) (bool, error) {
```
sethvargo, line 85: "Unsigned" and "breakglass" are used interchangeably in a few places, which might be confusing for future developers. We should standardize.

client-lib/go/client/jvs_client_test.go
capri-xiyue: Followed the logic here. https://github.com/abcxyz/jvs/blob/3b5e51097c09a19a4f2ecfff5be9c6ccb1fcdc3e/client-lib/go/client/jvs_client_test.go#L169
The return of `createBreakglassToken` is a jwt token.

client-lib/go/client/jvs_client.go
capri-xiyue: "Unsigned" and "breakglass" are not interchangeably in concept.
"unsigned" means tokens that are not signed.
"breakglass" must be unsigned and "breakglass" has justification category set to `breakglass`
I've standardize it in this PR.

client-lib/go/client/jvs_client_test.go
capri-xiyue: And we need both type `token` and type `string`

capri-xiyue: I've extracted such serialize logic into its own function to make it clear.

client-lib/go/client/jvs_client_test.go
yolocs: But we don't need to move the jwks server into subtest, do we?

client-lib/go/client/jvs_config.go
yolocs: To me, breakglass is a use case most services should and will adopt. That's why I think it should be allowed by default. Is your preference from being more strict by default?

client-lib/go/client/jvs_client_test.go
capri-xiyue: Sever is used in the jvs config for the jvs endpoint. I think it looks confusing that we initialize jvs client in subtests but the jwks server are shared and each subtests has its owner jwks server is better for test parallelism. Therefore, I moved all of them into subtests.

client-lib/go/client/jvs_client.go
```diff
@@ -57,5 +65,42 @@ func NewJVSClient(ctx context.Context, config *JVSConfig) (*JVSClient, error) {
 
 // ValidateJWT takes a jwt string, converts it to a JWT, and validates the signature.
 func (j *JVSClient) ValidateJWT(jwtStr string) (*jwt.Token, error) {
+	// Handle unsigned tokens.
+	if strings.HasSuffix(jwtStr, UnsignedPostfix) {
+		token, err := jwt.Parse([]byte(jwtStr), jwt.WithVerify(false))
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse jwt %s: %w", jwtStr, err)
+		}
+		valid, err := j.unsignedTokenValidAndAllowed(token)
```
yolocs, line 74: From this usage, it doesn't seem the `bool` return is necessary? Having an err is equivalent to having a false.

client-lib/go/client/jvs_client.go
yolocs: No fixed?

client-lib/go/client/jvs_client.go
yolocs: @raserva I wonder what's the meaning if not all justification has `breakglass` category?

client-lib/go/client/jvs_client_test.go
yolocs: Exactly because the jwks doesn't change, I recommend reusing the same server (not moving the server code into subtest).

client-lib/go/client/jvs_client_test.go
```diff
@@ -76,65 +76,56 @@ func TestValidateJWT(t *testing.T) {
 	if err != nil {
 		t.Fatal("couldn't create jwks json")
 	}
+	tok := testCreateToken(t, "test_id")
```
yolocs, line 79: I think you can inline these variable initialization to have cleaner code. Does the example below work?

client-lib/go/client/jvs_client_test.go
```diff
@@ -76,65 +76,56 @@ func TestValidateJWT(t *testing.T) {
...
 			name:    "unsigned",
 			jwt:     unsignedJWT,
-			wantErr: "required field \"signatures\" not present",
-		}, {
+			wantErr: "justification category is not breakglass, denying",
+		},
+		{
+			name:      "breakglass",
+			jwt:       breakglassJWT,
+			wantToken: breakglassTok,
```
yolocs, line 120:
```suggestion
			jwt:       testCreateUnsignedJWT(t, testCreateBreakglassToken(t, "test_id_3")),
			wantToken: testCreateBreakglassToken(t, "test_id_3"),
```

client-lib/go/client/jvs_client.go
raserva: The expectation is to ensure that this is intended to be a breakglass token, and not just an unsigned token

raserva: if we didn't have some sort of sentinel value, then any unsigned JWT would be treated as a breakglass token, even if it was not intended to be.

yolocs: My question is whether we should ensure **all** categories are `breakglass` or just **one-of** them.

raserva: This logic looks over all justifications and if any are breakglass, it return true. Only if it never finds one does it return false

raserva: here it returns nil if it finds one, but doesn't force all categories to be breakglass

yolocs: Ack. We need to document this behavior in the user manual.

client-lib/go/client/jvs_client.go
capri-xiyue: It is fixed

capri-xiyue: done

client-lib/go/client/jvs_client_test.go
capri-xiyue: I inline the variable initialization for variables that will be used only once.

client-lib/go/client/jvs_client.go
```diff
@@ -57,5 +65,43 @@ func NewJVSClient(ctx context.Context, config *JVSConfig) (*JVSClient, error) {
...
+	// Handle unsigned tokens.
+	if strings.HasSuffix(jwtStr, UnsignedPostfix) {
+		token, err := jwt.Parse([]byte(jwtStr), jwt.WithVerify(false))
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse jwt %s: %w", jwtStr, err)
+		}
+		err = j.unsignedTokenValidAndAllowed(token)
+		if err != nil {
+			return nil, fmt.Errorf("token unsigned and could not be validated: %w", err)
+		}
```
yolocs, line 77:
```suggestion
		if err := j.unsignedTokenValidAndAllowed(token); err != nil {
			return nil, fmt.Errorf("token unsigned and could not be validated: %w", err)
		}
```

client-lib/go/client/jvs_client.go
```diff
@@ -57,5 +65,43 @@ func NewJVSClient(ctx context.Context, config *JVSConfig) (*JVSClient, error) {
...
+		return fmt.Errorf("can't find 'justs' in claims, denying")
+	}
+	justsBytes, err := json.Marshal(justs)
+	if err != nil {
+		return fmt.Errorf("failed to marshal 'justs', denying")
+	}
+	var justifications []*jvsapis.Justification
+	if err = json.Unmarshal(justsBytes, &justifications); err != nil {
+		return fmt.Errorf("failed to unmarshal 'justs', denying")
+	}
```
yolocs, line 99:
```suggestion
	if err := json.Unmarshal(justsBytes, &justifications); err != nil {
		return fmt.Errorf("failed to unmarshal 'justs', denying")
	}
```

yolocs (commented):
Pending remaining thread on whether to default to allow breakglass. Other looks good.

client-lib/go/client/jvs_config.go
capri-xiyue: If it's common for services to allow breakglass, I think it's fine to make the default to be allow breakglass. Otherwise, it's safer to be more strict by default. @sethvargo Any ideas?

sethvargo: Breakglass is inherently a way to bypass security controls and therefore should be disabled by default. As a rule of thumb, users should always have to opt into doing something insecure, even if it's a standard.

capri-xiyue: done. Defaults to breakglass forbidden.

yolocs: (I don't think JVS is technically a security control. Authz is security control. JVS is privacy focused.)

The risk on the other side seems to be an oncaller requesting access at 3am only ends up finding out the service doesn't allow breakglass due to it being the default.
I can't say which risk is worse, especially when we will still audit log the breakglass access. Unable to debug in worse cases could stop business continuity.

I'm fine flipping the flag if you think the security risk outweighs the risk on the other side.

yolocs (approved):

PR #119, opened 2022-09-21
Switch to using a single JWT library and unify api package

### LINKED ISSUE #116: Two jwt libraries
### TL;DR

There are two jwt libraries in JVS. One defines `jwt.Token` as struct, and the other defines it as an interface. We should standardize on a single jwt library.

### Expected behavior

Single jwt library

### Observed behavior

Both github.com/golang-jwt/jwt and github.com/lestrrat-go/jwx/v2/jwt libraries.

### Environment Details

```markdown
N/A
```

### Additional information

_No response_

sethvargo:
The current implementation uses two JWT packages: github.com/golang-jwt/jwt and github.com/lestrrat-go/jwx/v2/jwt. These packages have very different APIs and developer surfaces. This PR standardizes everything on github.com/lestrrat-go/jwx/v2/jwt since we need to the jwx package anyway for jwk parsing and jws signing.

Because github.com/lestrrat-go/jwx/v2/jwt uses an interface to store claims, this PR introduces new top-level functions for getting/setting justifications on a token. This is significantly less error-prone than relying on "justs" to be populated in the JSON and enforces a structure. The downside is that it requires an extra option during parsing, but that is abstracted away into helpers. These helpers are important because the previous implementation had some extremely rough edges because slices are passed by reference. This caused data races and extremely-difficult-to-debug side-effects. The helpers make a copy of slices to ensure this doesn't happen.

One of the major downsides of the github.com/lestrrat-go/jwx/v2/jwt package is that there is no way to create "unsigned" tokens. github.com/golang-jwt/jwt has a "SigningString()" function for this purpose, but that does not exist in the new package. In retrospect, I'm not convinced that using the ".NOT_SIGNED" suffix is the right solution for breakglass, and I think we should explore using HMAC-signed keys instead. That being said, I was able to preserve backwards-compatibility with the existing implementation by signing the key, stripping off the signature, and appending ".NOT_SIGNED". This is less than ideal, but I think any refactors to the breakglass semantics should be done in a separate PR and discussion.

This PR also has some general code cleanup, such as being consistent with the name of the jvsapis/v0 package import.

Closes #116


apis/v0/jwt.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/lestrrat-go/jwx/v2/jwt"
+)
+
+const (
+	// jwtJustificationsKey is the key in the JWT where justifications are stored.
+	// Ideally this would be "justifications", but the RFC and various online
+	// resources recommend key names be as short as possible to keep the JWTs
+	// small. Akamai recommends less than 8 characters and Okta recommends less
+	// than 6.
+	jwtJustificationsKey string = "justs"
```
sethvargo, line 29: To discuss: I originally thought about having this public, but I didn't want to encourage people to use the `jwt.Token` outside of our helpers.

apis/v0/jwt.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+// WithTypedJustifications is an option for parsing JWTs that will convert
+// decode the [Justification] claims into the correct Go structure. If this is
+// not supplied, the claims will be "any" and future type assertions will fail.
+func WithTypedJustifications() jwt.ParseOption {
```
sethvargo, line 35: This is one of the biggest downsides to the lestrrat library. There are globals for registering decoding hooks, but they won't work because of how we have the packages structured.

apis/v0/jwt.go
```diff
@@ -0,0 +1,107 @@
...
+		return []*Justification{}, nil
+	}
+
+	typ, ok := raw.([]*Justification)
+	if !ok {
+		return nil, fmt.Errorf("found justifications, but was %T (expected %T)",
+			raw, []*Justification{})
+	}
+
+	// Make a copy of the slice so we don't modify the underlying data structure.
```
sethvargo, line 59: This is actually a fun bug in the previous implementation, since modifying the returned slice would actually modify the token. The performance overhead of making these copies seems work the security trade-off risks of setting invalid justifications.

apis/v0/jwt.go
```diff
@@ -0,0 +1,107 @@
...
+	cp = append(cp, justifications...)
+	return t.Set(jwtJustificationsKey, cp)
+}
+
+// AppendJustification appends the given justification to the end of the current
+// justifications list. It does not check for duplication and does not lock the
+// token. There is a possible race between when the claims are read and when the
+// claims are set back on the token. Callers should use [SetJustifications]
+// directly to avoid this race.
+func AppendJustification(t jwt.Token, justification *Justification) error {
```
sethvargo, line 82: To discuss: should we have this function? It has hairy edge cases (it's not safe for concurrent use), and I couldn't find a legitimate use case for it in the existing code.

pkg/cli/token.go
```diff
@@ -19,26 +19,33 @@ import (
...
+	Issuer  = "jvsctl"
+	Subject = "jvsctl"
+)
+
 var (
-	tokenExplanation string
-	breakglass       bool
-	ttl              time.Duration
-	issTimeUnix      int64
+	flagTokenExplanation string
```
sethvargo, line 45: This is a best practice we should try to codify somewhere. Flag names in CLIs should be prefixed with `flag` to differentiate them from other vars.

client-lib/go/client/jvs_config.go
```diff
@@ -32,8 +32,9 @@ type JVSConfig struct {
 	// Version is the version of the config.
 	Version string `yaml:"version,omitempty" env:"VERSION,overwrite,default=1"`
 
-	// JVS Endpoint. Expected to be fully qualified, including port. ex. http://127.0.0.1:8080
-	JVSEndpoint string `yaml:"endpoint,omitempty" env:"ENDPOINT,overwrite"`
+	// JWKSEndpoint is the full path (including protocol and port) to the JWKS
```
sethvargo, line 35: I renamed this to more accurately reflect the expected value and updated the comment/example.

yolocs (commented):
This is super to make things consistent! Immense thanks!

pkg/cli/token.go
```diff
@@ -139,30 +146,46 @@ func callOpts(ctx context.Context) ([]grpc.CallOption, error) {
 	return []grpc.CallOption{grpc.PerRPCCredentials(oauth.NewOauthAccess(token))}, nil
 }
 
+// breakglassToken creates a new breakglass token. The token is unsigned.
+//
+// TODO(sethvargo): should this be .NOT_SIGNED or should we sign the token with
+// HMAC using a shared secret exposed via the client.
```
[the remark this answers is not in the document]
yolocs: Saw the comment below

pkg/cli/token.go
```diff
@@ -139,30 +146,46 @@ func callOpts(ctx context.Context) ([]grpc.CallOption, error) {
...
+
+	if err := jvspb.SetJustifications(token, justs); err != nil {
+		return "", fmt.Errorf("failed to set justifications on jwt: %w", err)
 	}
 
-	return tok + ".NOT_SIGNED", nil
+	// TODO(sethvargo): the jwt library doesn't actually provide an API for
+	// exposing the raw signing string, so the easiest way to get it is to use the
+	// HMAC signer. We should seriously consider just using an HMAC signer with a
+	// shared secret instead of the hardcoded string.
```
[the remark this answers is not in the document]
sethvargo: We should talk about it in another PR/issue (I don't want to hold this PR up on making that decision, since this change is fully compat with the existing implementation). To support breakglass via HMAC, we'd inspect the JWS envelope and look at the `alg`. If it's `HS256`, we'd parse as breakglass and verify the HMAC. If it's `ES256`, we'd use the JWKs:

```go
// psuedo-code
envelope, err := jws.ParseString(tokenStr, jws.Validate(false))
if err != nil {
  panic(err)
}

var token jwt.Token
switch envelope.Algorithm() {
  case jws.HS256:
    token, err = jwt.ParseString(tokenStr, jws.WithKey(jws.HS256, []byte("jvs")))
  case jws.ES256:
    token, err = jwt.ParseString(tokenStr, jws.WithKeySet(jwksSet, jws.InferAlgorithmFromKey(true))
  default:
    panic("invalid algorithm")
}
```

Those aren't totally the APIs, but I think that gives you the idea.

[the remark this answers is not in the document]
yolocs: Created: https://github.com/abcxyz/jvs/issues/120. We can discuss it there.

apis/v0/jwt.go
```diff
@@ -0,0 +1,104 @@
...
+		return nil, fmt.Errorf("token cannot be nil")
+	}
+
+	raw, ok := t.Get(jwtJustificationsKey)
+	if !ok {
+		return []*Justification{}, nil
+	}
+
+	var claims []*Justification
+	switch list := raw.(type) {
```
yolocs, line 62: Nice

yolocs (approved):

PR #122, opened 2022-09-21
Switch to using HMAC for breakglass tokens

sethvargo:
POC, needs tests and also the Java implementation.

/cc @yolocs


apis/v0/breakglass.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/lestrrat-go/jwx/v2/jwa"
+	"github.com/lestrrat-go/jwx/v2/jws"
+	"github.com/lestrrat-go/jwx/v2/jwt"
+)
+
+const (
+	// BreakglassHMACSecret is the HMAC key to use for creating breakglass tokens.
+	// Breakglass tokens are already "unverified", so having this static secret
+	// does not introduce additional risk, and breakglass is disabled by default.
+	BreakglassHMACSecret = "BHzwNUbxcgpNoDfzwzt4Dr2nVXByUCWl1m8Eq2Jh26CGqu8IQ0VdiyjxnCtNahh9" //nolint:gosec
```
yolocs, line 29: I've also seen practice to name such variable as "XxxNotSoSecret"

yolocs: But I wonder why use such a complicated string instead of something like "BREAKGLASS"

client-lib/java/src/main/java/com/abcxyz/jvs/JvsClient.java
```diff
@@ -34,52 +36,80 @@
...
+    }
+
+    List<Justification> justifications = token.getClaim("justs").asList(Justification.class);
+    for (Justification justification : justifications) {
+      if (BREAKGLASS_JUSTIFICATION_CATEGORY.equals(justification.getCategory())) {
+        return token;
       }
     }
 
+    log.error("token is breakglass, but is missing breakglass justification");
```
yolocs, line 80: In Go, we will return error if no breakglass explanation is found. So should we throw an exception here?

client-lib/java/src/main/java/com/abcxyz/jvs/JvsConfiguration.java
```diff
@@ -35,7 +35,7 @@ public class JvsConfiguration {
   private String version = EXPECTED_VERSION;
 
   @JsonProperty("endpoint")
```
yolocs, line 37: It would be better if we update this config key as well to be `jwks_endpoint`

client-lib/java/src/main/java/com/abcxyz/jvs/Justification.java
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.abcxyz.jvs;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+
+@JsonDeserialize(using = JustificationDeserializer.class)
+public class Justification {
```
yolocs, line 22: This is fine. But I wonder if we _should_ use the proto generated `Justification` type instead?

yolocs (commented):
LGTM for the Go changes. @raserva for Java readability.

client-lib/java/src/main/java/com/abcxyz/jvs/Justification.java
raserva, line 22: This is a bog-standard Java data class. All the boilerplate should be replaced using [lombok](https://www.summittosound.com/services/foundation-inspections/). 

e.g.
```
@Data
public class Justification {
    private String category;
    private String value;
}
```

client-lib/java/src/main/java/com/abcxyz/jvs/JustificationDeserializer.java
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  public JustificationDeserializer(Class<?> vc) {
+    super(vc);
+  }
+
+  @Override
+  public Justification deserialize(JsonParser jp, DeserializationContext ctxt)
+      throws IOException, JsonProcessingException {
+    JsonNode node = jp.getCodec().readTree(jp);
```
raserva, line 38: null pointers possible here.

client-lib/java/src/main/java/com/abcxyz/jvs/JustificationDeserializer.java
```diff
@@ -0,0 +1,57 @@
...
+  public JustificationDeserializer() {
+    this(null);
+  }
+
+  public JustificationDeserializer(Class<?> vc) {
+    super(vc);
+  }
+
+  @Override
+  public Justification deserialize(JsonParser jp, DeserializationContext ctxt)
```
raserva, line 36: typically Java prefers verbosity. s/ctxt/context, s/jp/jsonParser.

client-lib/java/src/main/java/com/abcxyz/jvs/JustificationDeserializer.java
```diff
@@ -0,0 +1,57 @@
...
+package com.abcxyz.jvs;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import java.io.IOException;
+
+public class JustificationDeserializer extends StdDeserializer<Justification> {
```
raserva, line 26: Not clear to me why we're creating a custom deserializer rather than the standard object mapper. 

```
ObjectMapper objectMapper = new ObjectMapper();
Justification justification = objectMapper.readValue(json, Justification.class);
```

client-lib/java/src/main/java/com/abcxyz/jvs/JvsClient.java
```diff
@@ -34,52 +36,80 @@
...
   private final boolean allowBreakglass;
 
-  public DecodedJWT validateJWT(String jwtString) throws JwkException {
-    DecodedJWT jwt = JWT.decode(jwtString);
+  /**
+   * This parses the given token as a breakglass token. If the token is not a JWT, not HS256, or
+   * missing breakglass justifications, it returns null. If the token fails to verify, it throws an
+   * exception.
+   */
+  private DecodedJWT parseBreakglassToken(String tokenStr) throws JwkException {
```
raserva, line 56: So, this is a bit of a nit, but the practice of returning null as a "fail" is not super prevalent within java. With the name of this method, and the fact it shows that it throws a "JwkException" i would expect that i would get that exception if i passed this a non breakglass token. 

The more idiomatic way, in my mind, is to move the breakglass check to the validateJWT method, and only call parseBreakglassToken it it passes. 
(specifically the below moved) 

```
    if (!Header.JWT_TYPE.equals(token.getType())
        || !SignatureAlgorithm.HS256.getValue().equals(token.getAlgorithm())) {
      log.debug("token is not breakglass");
      return null;
    }
```

raserva: oh, i see it used to be similiar to that with a separate method to check that it was valid and allowed. Why did we refactor that out?

yolocs: I think the logic here doesn't regard a non-breakglass token as an error. It's like "not breakglass I don't care"

raserva: Well, given the name is "parseBreakglass" i would expect an error if i passed it a non breakglass token.

client-lib/java/src/main/java/com/abcxyz/jvs/Justification.java
sethvargo: I'm not super convinced we should be embedding the protos directly in Go either. I think we should have auxiliary structures here.

sethvargo: Done

client-lib/java/src/main/java/com/abcxyz/jvs/JustificationDeserializer.java
sethvargo: Two reasons:

1. I didn't know about `ObjectMapper`. The existing implementation manually creates a HashMap and isn't tied to structured values.
2. I prefer an interim data structure in case we need to mutate fields in the future for backwards compatability.

client-lib/java/src/main/java/com/abcxyz/jvs/JustificationDeserializer.java
sethvargo: Done.

client-lib/java/src/main/java/com/abcxyz/jvs/JustificationDeserializer.java
sethvargo: Done.

client-lib/java/src/main/java/com/abcxyz/jvs/JvsClient.java
sethvargo: I switched to to `validateBreakglassToken` and moved the header check into validateJWT.

client-lib/java/src/main/java/com/abcxyz/jvs/JvsClient.java
sethvargo: Updated

client-lib/java/src/main/java/com/abcxyz/jvs/JvsConfiguration.java
sethvargo: Done.

apis/v0/breakglass.go
sethvargo: Technically we need at least 64 bytes to avoid leaking. Go doesn't enforce this, but Java does (it throws an exception if you give the HMAC256 less than 64 bytes of key material). We could opt to make it a really long human-readable string, but it would need to be 64 bytes or longer.

client-lib/java/src/main/java/com/abcxyz/jvs/JustificationDeserializer.java
raserva: ObjectMapper is the standard solution in Java. I'm hesitant to go with a custom deserializer when we don't have a strong reason to have one.

apis/v0/breakglass.go
yolocs: LOL. Made one but up to you `ABCXYZ-JVS-BREAKGLASS-TOKEN-HMAC-NOT-SO-SECRET-XXXXXXXXXXXXXXXXX`

client-lib/java/src/main/java/com/abcxyz/jvs/Justification.java
yolocs: (Not blocking)
Yea, you mentioned it in an earlier PR. I wonder what's the purpose of having an auxiliary structure?

yolocs (approved):

client-lib/java/src/main/java/com/abcxyz/jvs/JustificationDeserializer.java
sethvargo: I guess I'm confused why we didn't use `ObjectMapper` in the first place then?

apis/v0/breakglass.go
sethvargo: Lol let's stick to the random string.

client-lib/java/src/main/java/com/abcxyz/jvs/JustificationDeserializer.java
raserva: Looks like we were converting it to a map, which already had a built in deserializer. ObjectMapper gets you custom objects easily, which we're doing now.

sethvargo: Chatted with Ryan offline. The JWT library makes using ObjectMapper prohibitively difficult and is less readable.

raserva: Talked with @sethvargo offline. It appears there is no easy way to get the string json representation of the "justs" claim to convert using the Object Mapper. Rather than use convoluted code to serialize to a string just to deserialize to Justifications, we can stick with this. The major thing i dislike is that now both the object and the deserializer have to be modified if there are new values.

I'm now wondering if it makes sense to except (or at least, log to keep backwards compatibility) on unhandled values

Issue #22, opened 2022-05-05
Add Additional detail to claims


raserva:
our claims have some fields as TODO. Those should be added with real values.

yolocs:
Mainly `aud` claim we have to fill

PR #123, opened 2022-09-23
feat: allow audiences for justification

yolocs:
Fix: https://github.com/abcxyz/jvs/issues/22


pkg/justification/processor.go
```diff
@@ -63,7 +63,8 @@ func NewProcessor(kms *kms.KeyManagementClient, config *config.JustificationConf
 }
 
 const (
-	cacheKey = "signer"
+	cacheKey        = "signer"
+	DefaultAudience = "jvs"
```
sethvargo, line 67: These are usually URLs or reverse DNS like `com.abcxyz.jvs`

myurtoglu (approved):

pkg/justification/processor.go
```diff
@@ -163,8 +164,14 @@ func (p *Processor) createToken(ctx context.Context, now time.Time, req *jvspb.C
 	exp := now.Add(req.Ttl.AsDuration())
 	justs := req.Justifications
 
+	// Use audiences in the request if provided.
+	aud := []string{DefaultAudience}
```
sethvargo, line 168: I might just be me, but this feels weird to alloc the aud that you might not use. Consider:

```go
aud := req.Audiences
if len(aud) == 0 {
  aud = []string{DefaultAudience}
}
```

protos/v0/jvs_request.proto
```diff
@@ -25,6 +25,9 @@ option go_package = "github.com/abcxyz/jvs/apis/v0";
 message CreateJustificationRequest {
   repeated Justification justifications = 1;
   google.protobuf.Duration ttl = 2;
+
+  // Optional audiences for the justifications.
+  repeated string audiences = 3;
```
sethvargo, line 30: Do you need to regen the protos?

yolocs: Yes, did that in the PR.

pkg/justification/processor.go
```diff
@@ -63,7 +63,8 @@ func NewProcessor(kms *kms.KeyManagementClient, config *config.JustificationConf
 }
 
 const (
-	cacheKey = "signer"
+	cacheKey        = "signer"
+	DefaultAudience = "dev.abcxyz.jvs"
```
sethvargo, line 67: Nit: document

sethvargo (approved):

Issue #104, opened 2022-08-02
Refactor of config pkg


capri-xiyue:
### TL;DR

There are some duplicates when loading config from env, do validation and set defaults.
https://github.com/abcxyz/jvs/blob/main/pkg/config/public_key_config.go#L43 and https://github.com/abcxyz/jvs/blob/23cdda185a8fc277ba798bfaf3e34543458f1934/pkg/config/crypto_config.go#L102.

We can do refactor to avoid duplicates.

### Detailed design

Create a common function which handles loading config from env with `any` parameter. The caller of the common function need to pass the pointer of specific config like `PublicKeyConfig` or ` CryptoConfig` as the value of `any` parameter.

yolocs:
Added a lib in pkg: https://github.com/abcxyz/pkg/pull/33

yolocs:
Adopt pkg/cfgloader wherever applicable

PR #125, opened 2022-09-23
cleanup: use common config loader to simplify code

yolocs:
Fix: https://github.com/abcxyz/jvs/issues/104


cmd/cert-rotation/main.go
```diff
@@ -83,8 +84,9 @@ func realMain(ctx context.Context) error {
 	}
 	defer kmsClient.Close()
 
-	config, err := config.LoadCryptoConfig(ctx, []byte{})
-	if err != nil {
+	// TODO(#124): We shouldn't need JVS_ prefix since it's a JVS service.
+	config := &config.CryptoConfig{}
```
sethvargo, line 88: Two nits:

1. The variable is shadowing the package name (both `config`). Consider `cfg` instead.
2. `var cfg Config` and then `&cfg` is usually more common.

(Throughout, not going to comment everywhere).

sethvargo (approved):

myurtoglu (approved):

Issue #112, opened 2022-08-09
Add a CLI command to validate jvs token


yolocs:
This will be handy for debug or manually validate token

PR #128, opened 2022-09-26
feat: add validate CLI

sqin2019:
to address #112

### DIFF
--- a/docs/jvs-quick-setup.md
+++ b/docs/jvs-quick-setup.md
@@ -79,7 +79,7 @@ will also get created.
 
     ```shell
     export SERVER=<jvs_server_domain>:443
-    export JWKS_ENDPOINT=public_key_server_url
+    export JWKS_ENDPOINT=public_key_server_url/.well-known/jwks
     ```
 
 2.  Create Justification Token via [jvsctl](cli-tool.md):
--- a/docs/jvs-quick-setup.md
+++ b/docs/jvs-quick-setup.md
@@ -79,7 +79,7 @@ will also get created.
 
     ```shell
     export SERVER=<jvs_server_domain>:443
-    export JWKS_ENDPOINT=public_key_server_url/.well-known/jwks
+    export JWKS_ENDPOINT=<public_key_server_url>/.well-known/jwks
     ```
 
 2.  Create Justification Token via [jvsctl](cli-tool.md):
--- a/docs/jvs-quick-setup.md
+++ b/docs/jvs-quick-setup.md
@@ -72,11 +72,12 @@ will also get created.
 
 ### Justification API
 
-1.  Export the jvs server with the domain part of the `jvs_server_url` from
+1.  Export the jvs server and JWKS endpoint with the domain part of the `jvs_server_url` from
     Terraform outputs like `jvs-e2e-xxxx-uc.a.run.app`
 
     ```shell
     export SERVER=<jvs_server_domain>:443
+    export JWKS_ENDPOINT=<jvs_server_domain>:8080/.well-known/jwks
     ```
 
 2.  Create Justification Token via [jvsctl](cli-tool.md):
@@ -85,8 +86,11 @@ will also get created.
     jvsctl token --explanation "issues/12345" --ttl 30m --server ${SERVER}
     ```
 
-**TODO(#112):** Once we have the "validate token" command, we can even validate
-it.
+3.  Validate Justification [jvsctl](cli-tool.md):
+
+    ```shell
+    jvsctl validate --token "example token" --jwks_endpoint ${JWKS_ENDPOINT}
+    ```
 
 ### Public Key API
 
--- a/docs/jvs-quick-setup.md
+++ b/docs/jvs-quick-setup.md
@@ -72,12 +72,14 @@ will also get created.
 
 ### Justification API
 
-1.  Export the jvs server and JWKS endpoint with the domain part of the `jvs_server_url` from
-    Terraform outputs like `jvs-e2e-xxxx-uc.a.run.app`
+1.  Export the jvs server with the domain part of the `jvs_server_url` from
+    Terraform outputs like `jvs-e2e-xxxx-uc.a.run.app`, and export the JWKS
+    endpoint with `public_key_server_url` from Terraform outputs if you want
+    to validate token via CLI.
 
     ```shell
     export SERVER=<jvs_server_domain>:443
-    export JWKS_ENDPOINT=<jvs_server_domain>:8080/.well-known/jwks
+    export JWKS_ENDPOINT=public_key_server_url
     ```
 
 2.  Create Justification Token via [jvsctl](cli-tool.md):
@@ -90,6 +92,10 @@ will also get created.
 
     ```shell
     jvsctl validate --token "example token" --jwks_endpoint ${JWKS_ENDPOINT}
+
+    # or pass token via pipe
+    echo "${JVS_TOKEN}" | jvsctl validate --token -
+    cat /tmp/jvs_token | jvsctl validate --token -
     ```
 
 ### Public Key API
--- a/pkg/cli/root_test.go
+++ b/pkg/cli/root_test.go
@@ -27,8 +27,9 @@ import (
 func TestInitCfg(t *testing.T) {
 	cfgFile = filepath.Join(t.TempDir(), ".jvsctl.yaml")
 
-	if err := os.WriteFile(cfgFile, []byte(`server: https://example.com
-`), fs.ModePerm); err != nil {
+	if err := os.WriteFile(cfgFile, []byte(
+		`server: https://example.com
+jwks_endpoint: https://jvs.corp:8080/.well-known/jwks`), fs.ModePerm); err != nil {
 		t.Fatalf("failed to prepare test config file: %v", err)
 	}
 	t.Cleanup(func() {
@@ -43,6 +44,7 @@ func TestInitCfg(t *testing.T) {
 		Version:        "1",
 		Server:         "https://example.com",
 		Authentication: &config.CLIAuthentication{},
+		JWKSEndpoint:   "https://jvs.corp:8080/.well-known/jwks",
 	}
 	if diff := cmp.Diff(wantCfg, cfg); diff != "" {
 		t.Errorf("CLI config loaded (-want,+got):\n%s", diff)
--- a/pkg/cli/root_test.go
+++ b/pkg/cli/root_test.go
@@ -29,7 +29,7 @@ func TestInitCfg(t *testing.T) {
 
 	if err := os.WriteFile(cfgFile, []byte(
 		`server: https://example.com
-jwks_endpoint: https://jvs.corp:8080/.well-known/jwks`), fs.ModePerm); err != nil {
+jwks_endpoint: https://example.com/.well-known/jwks`), fs.ModePerm); err != nil {
 		t.Fatalf("failed to prepare test config file: %v", err)
 	}
 	t.Cleanup(func() {
@@ -44,7 +44,7 @@ jwks_endpoint: https://jvs.corp:8080/.well-known/jwks`), fs.ModePerm); err != ni
 		Version:        "1",
 		Server:         "https://example.com",
 		Authentication: &config.CLIAuthentication{},
-		JWKSEndpoint:   "https://jvs.corp:8080/.well-known/jwks",
+		JWKSEndpoint:   "https://example.com/.well-known/jwks",
 	}
 	if diff := cmp.Diff(wantCfg, cfg); diff != "" {
 		t.Errorf("CLI config loaded (-want,+got):\n%s", diff)
--- a/pkg/cli/validate.go
+++ b/pkg/cli/validate.go
@@ -15,16 +15,22 @@
 package cli
 
 import (
+	"bufio"
 	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"sort"
+	"text/tabwriter"
 	"time"
 
+	jvspb "github.com/abcxyz/jvs/apis/v0"
 	"github.com/abcxyz/jvs/client-lib/go/client"
 	"github.com/spf13/cobra"
 )
 
 var (
-	flagToken           string
-	flagAllowBreakglass bool
+	flagToken string
 )
 
 var validateCmd = &cobra.Command{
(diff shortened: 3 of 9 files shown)


docs/jvs-quick-setup.md
```diff
@@ -72,11 +72,12 @@ will also get created.
 
 ### Justification API
 
-1.  Export the jvs server with the domain part of the `jvs_server_url` from
+1.  Export the jvs server and JWKS endpoint with the domain part of the `jvs_server_url` from
     Terraform outputs like `jvs-e2e-xxxx-uc.a.run.app`
 
     ```shell
     export SERVER=<jvs_server_domain>:443
+    export JWKS_ENDPOINT=<jvs_server_domain>:8080/.well-known/jwks
```
sqin2019, line 80: @yolocs Could you confirm my understanding is correct here? thx!

yolocs: The jwks endpoint could be different from the JVS API server. But it's definitely preferable to put them behind the same GCLB and share a custom domain. I think we can do two things:

1. If the jwks endpoint is not specified in the config file, default the jwks endpoint to be `https://{SERVER}/.well-known/jwks`. We probably can't use viper for such defaulting. We need to handle that explicitly in code.

2. Allow overwriting the jwks endpoint via config/flag.

pkg/cli/root_test.go
```diff
@@ -27,8 +27,9 @@ import (
 func TestInitCfg(t *testing.T) {
 	cfgFile = filepath.Join(t.TempDir(), ".jvsctl.yaml")
 
-	if err := os.WriteFile(cfgFile, []byte(`server: https://example.com
-`), fs.ModePerm); err != nil {
+	if err := os.WriteFile(cfgFile, []byte(
+		`server: https://example.com
+jwks_endpoint: https://jvs.corp:8080/.well-known/jwks`), fs.ModePerm); err != nil {
```
yolocs, line 32:
```suggestion
jwks_endpoint: https://example.com/.well-known/jwks`), fs.ModePerm); err != nil {
```

pkg/cli/validate.go
```diff
@@ -0,0 +1,63 @@
...
+	"context"
+	"time"
+
+	"github.com/abcxyz/jvs/client-lib/go/client"
+	"github.com/spf13/cobra"
+)
+
+var (
+	flagToken           string
+	flagAllowBreakglass bool
```
yolocs, line 27: I wonder if we need this to be a flag. We should always allow breakglass but print out a warning if it's a breakglass token.

We can check if it's a breakglass token with https://github.com/abcxyz/jvs/blob/main/apis/v0/breakglass.go#L64

Alternatively, we could use such a flag to control whether to return an error but we should still parse the breakglass token and output its content.

pkg/cli/validate.go
```diff
@@ -0,0 +1,63 @@
...
+		JWKSEndpoint:    cfg.JWKSEndpoint,
+		CacheTimeout:    5 * time.Minute,
+		AllowBreakglass: flagAllowBreakglass,
+	})
+	if err != nil {
+		return err
+	}
+
+	_, err = jvsclient.ValidateJWT(flagToken)
+	if err != nil {
```
yolocs, line 51:
```suggestion
	if _, err := jvsclient.ValidateJWT(flagToken); err != nil {
```

pkg/cli/validate.go
```diff
@@ -0,0 +1,63 @@
...
+		return err
+	}
+
+	_, err = cmd.OutOrStdout().Write([]byte("Token is valid"))
```
yolocs, line 55: We should print out the parsed token content instead?

pkg/cli/validate_test.go
```diff
@@ -0,0 +1,162 @@
...
+	"crypto/elliptic"
+	"crypto/rand"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	v0 "github.com/abcxyz/jvs/apis/v0"
```
yolocs, line 28:
```suggestion
	jvspb "github.com/abcxyz/jvs/apis/v0"
```

pkg/cli/validate_test.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/abcxyz/jvs/pkg/config"
+	"github.com/abcxyz/pkg/testutil"
+	"github.com/lestrrat-go/jwx/v2/jwa"
+	"github.com/lestrrat-go/jwx/v2/jwk"
+	"github.com/lestrrat-go/jwx/v2/jws"
+	"github.com/lestrrat-go/jwx/v2/jwt"
+	"github.com/spf13/cobra"
+)
+
+func TestRunValidateCmd(t *testing.T) {
```
yolocs, line 38: `t.Parallel()`

pkg/cli/validate_test.go
```diff
@@ -0,0 +1,162 @@
...
+			allowBreakglass: false,
+		},
+		{
+			name:    "invalid",
+			token:   "token",
+			wantErr: "failed to parse token headers",
+		},
+	}
+
+	for _, tc := range tests {
```
yolocs, line 113: IIUC, the tests can't paralleled because we are setting global flags? If so, let's add a comment explaining that.

pkg/config/cli_config.go
```diff
@@ -32,6 +32,10 @@ type CLIConfig struct {
 
 	// Authentication is the authentication config.
 	Authentication *CLIAuthentication `yaml:"authentication,omitempty"`
+
+	// JWKSEndpoint is the full path (including protocol and port) to the JWKS
+	// endpoint on a JVS server (e.g. https://jvs.corp:8080/.well-known/jwks).
+	JWKSEndpoint string `yaml:"jwks_endpoint,omitempty" mapstructure:"jwks_endpoint"`
```
yolocs, line 38: Please update the comment per my comment above

yolocs (changes requested):

docs/jvs-quick-setup.md
sethvargo: I'm not sure if Viper supports it, but we could have JWS_ENDPOINT default to `${SERVER}:8080/.well-known/jwks`

pkg/cli/root_test.go
sethvargo: This is probably my fault - I use `jvs.corp` in the godoc comment.

pkg/cli/validate.go
sethvargo: I think I would expect `jvsctl validate` to print out all information about the token, including whether it's breakglass. It should exit 1 if validation fails and exit 0 if validation succeeds. But either way, I'd expect to see `breakglass: true|false` in the output.

pkg/cli/validate.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/abcxyz/jvs/client-lib/go/client"
+	"github.com/spf13/cobra"
+)
+
+var (
+	flagToken           string
```
sethvargo, line 26: I'm a little bit worried about supplying the token as a string, since then the JWT will appear in shell history. I wonder if we should support reading this from a pipe (e.g. stdin) or a raw string. For example:

```shell
jvsctl validate -token="ey2..."
```

would work as-is, but:

```shell
jvsctl validate -token=-
```

would read from stdin. That would allow callers to read the token from a file or envvar:

```shell
echo "${JVS_TOKEN}" | jvsctl validate -token=-
cat /tmp/jvs_token | jvsctl validate -token=-
```

pkg/cli/validate.go
```diff
@@ -0,0 +1,63 @@
...
+	ctx := context.Background()
+
+	jvsclient, err := client.NewJVSClient(ctx, &client.JVSConfig{
+		Version:         "1",
+		JWKSEndpoint:    cfg.JWKSEndpoint,
+		CacheTimeout:    5 * time.Minute,
+		AllowBreakglass: flagAllowBreakglass,
+	})
+	if err != nil {
+		return err
```
sethvargo, line 47:
```suggestion
		return fmt.Errorf("failed to create jvs client: %w", err)
```

pkg/cli/validate.go
sethvargo: +1. I thought we had a linter for this, but apparently it was deprecated :(

pkg/cli/validate.go
sethvargo: +1 - I left a similar comment above. Here's how I'm thinking about possible outputs:

**Completely invalid token**

```text
failed to parse token: not a valid jwt
(exit 65)
```

**Valid token but invalid signature**

```text
audiences     foo.bar, zip.zap
breakglass    false
subject       https://service.corp
# ...
valid         false
(exit 74)
```

**Valid breakglass token**

```text
audiences     foo.bar, zip.zap
breakglass    true
subject       https://service.corp
# ...
valid         true
(exit 0)
```

**Valid token**

```text
audiences     foo.bar, zip.zap
breakglass    false
subject       https://service.corp
# ...
valid         true
(exit 0)
```

sethvargo (commented):
I left a few inline comments, but my biggest piece of feedback is around the output formatting and handling.

1. I think we need to consider supporting multiple output formats (e.g. table for humans, json for machines).
2. If the token is parseable, we should print everything we know about it.
3. If the token is unparseable OR invalid signature, exit non-zero.

It might make sense to have general-purpose functions in `jvspb` for printing a JWT as a table or JSON.

docs/jvs-quick-setup.md
sqin2019: Ack.

yolocs: Minor correction. `SERVER` is the gRPC server address. It will be in the format of e.g. `example.com:443`. Without scheme `https://` and with port.
The JWT endpoint should be an http URL.

pkg/cli/validate.go
yolocs: +1 I forgot to comment on this. In addition, I think the token should be an argument instead of a flag. `echo "${JVS_TOKEN}" | jvsctl validate -`

pkg/cli/validate.go
yolocs: Any lib recommendation? Things like https://github.com/olekukonko/tablewriter might be overkill.

pkg/cli/validate_test.go
sqin2019: Could you clarify they "cannot" or "should not" run parallely with global flags? I am able to run and pass the tests with t.Parallel() therefore wondering.

pkg/cli/validate.go
sethvargo: Go has a tabwriter package in stdlib iirc. However, it's not a ton of code to write ourselves:

```go
claimsMap, ok := token.AsMap(ctx)
if !ok {
  # ...
}

longest := 0
claims := make(map[string]string, len(claimsMap))
claimsKeys := make([]string, 0, len(claimsMap))
for k, v := range claimsMap {
  if l := len(k); l > longest {
    longest = l
  }
  claimsKeys = append(claimsKeys, k)
  claims[k] = toString(v)
}
sort.Strings(claimsKeys)

spacing := longest + 5 // arbitrary
for _, k := range claimsKeys {
  fmt.Fprint(w, k)
  for i := 0; i < longest-len(k); i++ {
    fmt.Fprint(w, " ")
  }
  fmt.Fprint(w, claims[k])
  fmt.Fprintln(w)
}
```

I'm sure there's a bounds error or something in there, but that's the general idea.

pkg/cli/validate.go
sethvargo: I slightly prefer flags, because I prefer being explicit, but I'm largely indifferent.

pkg/cli/validate.go
yolocs: I'm absolute fine to start with something primitive.

docs/jvs-quick-setup.md
sqin2019: To clarify, they should look like below?
```
SERVER=<jvs_server_domain>:443
Default JWKS_ENDPOINT=https://<jvs_server_domain>:8080/.well-known/jwks
```

pkg/cli/validate_test.go
sethvargo:
```
go test -count=1 -race -shuffle=on -parallel=20
```

docs/jvs-quick-setup.md
sqin2019: Updated, please take a look.

pkg/cli/validate.go
```diff
@@ -41,23 +47,84 @@ func runValidateCmd(cmd *cobra.Command, args []string) error {
...
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create jvs client: %w", err)
 	}
 
-	_, err = jvsclient.ValidateJWT(flagToken)
+	if flagToken == "-" {
+		stat, _ := os.Stdin.Stat()
+		if (stat.Mode() & os.ModeCharDevice) == 0 {
+			// Read from pipe
```
sethvargo, line 59: It might be shorter to read all of stdin:


```go
b, err := ioutil.ReadAll(io.LimitReader(stdin, 64*1_000))
```

Also, I would recommend making `os.Stdin` a private global variable. It makes overriding in tests easier.

pkg/cli/validate.go
```diff
@@ -41,23 +47,84 @@ func runValidateCmd(cmd *cobra.Command, args []string) error {
...
+			return err
+		}
+	}
+	w.Flush()
+	return nil
+}
 
-	_, err = cmd.OutOrStdout().Write([]byte("Token is valid"))
-	return err
+func toString(v interface{}) {
```
sethvargo, line 123: ?

sqin2019: will remove...

pkg/cli/validate.go
sqin2019: Used io.ReadAll since ioutil.ReadAll is deprecated.

sethvargo: Sorry - old habits :)

pkg/cli/validate.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// Output token claims into a table
+	w := tabwriter.NewWriter(cmd.OutOrStdout(), 0, 0, 1, ' ', 0)
+	for _, k := range claimsKeys {
+		if _, err := fmt.Fprintf(w, "%s\t%s", k, claims[k]); err != nil {
+			return err
+		}
+		fmt.Fprintln(w)
+	}
+	w.Flush()
```
yolocs, line 117: I start to doubt the value of such a table output. Wouldn't yaml encoding easier for reading, especially nested values? In the current code, we're encoding nested values to be json string again. See my example: https://go.dev/play/p/-zpkCXBjayn
cc: @sethvargo

pkg/cli/validate.go
```diff
@@ -0,0 +1,124 @@
...
+	return nil
+}
+
+func init() {
+	validateCmd.Flags().StringVarP(&flagToken, "token", "t", "", "The token that needs validation")
+	validateCmd.MarkFlagRequired("token") //nolint // not expect err
```
yolocs, line 123: Shall we have a flag for the output format? By default, it's "table" (or "yaml" if you take my suggestion in the other comment) or "json".

yolocs (changes requested):

pkg/cli/validate.go
yolocs: Had an offline discussion. The only nested claim seems to be `justs` and due the nature of this being a JVS CLI, this claim is important enough to be visually separated from standard claims. Using yaml output is optional.

E.g.

```
valid:		true
breakglass:	false
---
aud: ["jvs.abcxyz.dev"]
sub: "foo@example.com"
---
justifications:
	explanation: "Here is my reason"
	explanation: "Here is my other reason"
```

```
valid:		true
breakglass:	true
---
aud: ["jvs.abcxyz.dev"]
sub: "jvsctl"
---
justifications:
	breakglass: "Here is my breakglass reason"
```

pkg/cli/validate.go
yolocs: I'm fine skipping the `json` output for now.

pkg/cli/validate.go
sqin2019: Thanks for the feedback, on a second thought, should this entire "convert the token claims in yaml/json/table" (including what justs should be look like in there) be part of jvspb as general functions as Seth suggested earlier? Let me know, I will create a follow up task(TODO here).

yolocs: Another way: You can start a new PR to add that in `jvspb` first and after its merge update this PR to use it.

sqin2019:
To not confuse review with rebasing, opened a different PR, closing this one.

PR #132, opened 2022-09-28
Make JWKs generation deterministic

sethvargo:
The jwk.Set library sorts keys by their key material, which leads to some non-determinism in tests. We want to ensure keys are always sorted by key version name, so use our own data structure.


yolocs (approved):
👍

PR #134, opened 2022-09-29
Drop global variables in CLI

### LINKED ISSUE #133: jvsctl requires a server for breakglass
### TL;DR

The `jvsctl` command requires a `server` property, even for breakglass tokens.

### Expected behavior

Breakglass should not require a server.

### Observed behavior

Breakglass requires a server.

### Environment Details

```markdown
Everywhere
```

### Additional information

_No response_

sethvargo:
This removes the use of global variables for state in pkg/cli, which has the added benefit of allowing CLI tests to run in parallel. This new pattern ensures configuration is loaded and passed to child commands.

Fixes #133

### DIFF
--- a/pkg/cli/root.go
+++ b/pkg/cli/root.go
@@ -12,89 +12,83 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package cli implements the commands for the JVS CLI.
 package cli
 
 import (
 	"errors"
 	"fmt"
 	"os"
-	"path/filepath"
-
-	"github.com/spf13/cobra"
-	"github.com/spf13/viper"
 
 	"github.com/abcxyz/jvs/pkg/config"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
-var (
-	cfgFile string
-	cfg     *config.CLIConfig
-)
-
-var rootCmd = &cobra.Command{
-	Use:               "jvsctl",
-	Short:             "jvsctl facilitates the justification verification flow provided by abcxyz/jvs",
-	PersistentPreRunE: ensureCfg,
+// rootCmdOptions are used as options to the root command.
+type rootCmdOptions struct {
+	configPath string
 }
 
-// Execute executes the CLI.
-func Execute() error {
-	return rootCmd.Execute()
-}
+// newRootCmd creates a new instance of the root cobra command node.
+func newRootCmd(cfg *config.CLIConfig) *cobra.Command {
+	vpr := viper.New()
 
-func init() {
-	cobra.OnInitialize(initCfg)
-	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.jvsctl/config.yaml)")
-	rootCmd.PersistentFlags().String("server", "", "overwrite the JVS server address")
-	rootCmd.PersistentFlags().Bool("insecure", false, "use insecure connection to JVS server")
-	viper.BindPFlag("server", rootCmd.PersistentFlags().Lookup("server"))     //nolint // not expect err
-	viper.BindPFlag("insecure", rootCmd.PersistentFlags().Lookup("insecure")) //nolint // not expect err
+	opts := &rootCmdOptions{}
 
-	rootCmd.AddCommand(tokenCmd)
-}
+	cmd := &cobra.Command{
+		Use:   "jvsctl",
+		Short: "jvsctl facilitates the justification verification flow provided by abcxyz/jvs",
 
-func initCfg() {
-	if cfgFile != "" {
-		// Use config file from the flag.
-		viper.SetConfigFile(cfgFile)
-	} else {
-		// Try use the default config file.
-		home, err := os.UserHomeDir()
-		cobra.CheckErr(err)
-
-		viper.AddConfigPath(filepath.Join(home, ".jvsctl"))
-		viper.SetConfigType("yaml")
-		viper.SetConfigName("config")
+		// Load the configuration before each child command. The actual "compiled"
+		// configuration is passed in to the newChildCmd() function, but this sets
+		// the values of that config.
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			return loadConfig(vpr, opts.configPath, cfg)
+		},
 	}
 
-	// Also load from env vars.
-	viper.SetEnvPrefix("JVSCTL")
-	viper.AutomaticEnv()
+	// Flags
+	cmd.PersistentFlags().StringVarP(&opts.configPath, "config", "c", "",
+		"path to a file on disk for the jvs configuration")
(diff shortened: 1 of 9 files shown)


pkg/config/cli_config.go
```diff
@@ -30,13 +30,8 @@ type CLIConfig struct {
 	// Server is the JVS server address.
 	Server string `yaml:"server,omitempty"`
 
-	// Authentication is the authentication config.
```
sethvargo, line 33: ⚠️ This is a change. I don't know why this was an embedded struct.

pkg/config/cli_config.go
```diff
@@ -51,10 +46,6 @@ func (cfg *CLIConfig) Validate() error {
 			cfg.Version, CLIConfigVersions.List()))
 	}
 
-	if cfg.Server == "" {
```
sethvargo, line 54: This validation makes it so we cannot generate breakglass tokens.

pkg/cli/root.go
```diff
@@ -12,89 +12,83 @@
...
+	vpr.SetConfigType("yaml")
+	vpr.SetEnvPrefix("JVSCTL")
+	vpr.AutomaticEnv()
+	vpr.SetConfigFile(pth)
+
+	if err := vpr.ReadInConfig(); err != nil {
+		// Don't throw an error for failing to find the default config path since
+		// it's not required. However, if the user specified a config path that
+		// doesn't exist, that's an error.
+		if !errors.Is(err, &viper.ConfigFileNotFoundError{}) && !errors.Is(err, os.ErrNotExist) {
```
yolocs, line 78: I thought viper returns ConfigFileNotFoundError if os.ErrNotExist. Are you doing this for safety?

pkg/cli/root.go
```diff
@@ -12,89 +12,83 @@
...
+			return fmt.Errorf("failed to read configuration file: %w", err)
+		}
+		if pth != defaultConfigPath {
+			return fmt.Errorf("failed to read configuration file: %w", err)
+		}
```
yolocs, line 83: 👍

pkg/cli/root_test.go
```diff
@@ -15,36 +15,127 @@
...
+// testExecteCommand executes the given cobra command and returns the stdout,
+// stderr, and any execution error. See [testExecuteCommandStdin] for more
+// information.
+func testExecuteCommand(tb testing.TB, cmd *cobra.Command, args ...string) (string, string, error) {
+	tb.Helper()
+	return testExecuteCommandStdin(tb, cmd, nil, args...)
+}
+
+// testExecuteCommandStdin executes the given cobra command with the stdin
+// writer w and returns the stdout, stderr, and any errors that occur during
```
yolocs, line 123: "stdin reader"?

pkg/cli/token.go
```diff
@@ -19,64 +19,101 @@ import (
...
 	"github.com/abcxyz/jvs/pkg/idtoken"
 	"github.com/abcxyz/jvs/pkg/justification"
 )
 
-const (
-	Issuer  = "jvsctl"
-	Subject = "jvsctl"
-)
+// tokenCmdOptions holds all the inputs and flags for the token subcommand.
+type tokenCmdOptions struct {
```
yolocs, line 41: This is a good pattern to codify.

pkg/cli/token_test.go
```diff
@@ -15,190 +15,176 @@
...
+	b, err := jwt.Sign(token, jwt.WithKey(jwa.HS256, []byte("testing")))
 	if err != nil {
-		t.Fatal(err)
-	}
-	expectedJustifications := []*jvspb.Justification{
-		{
-			Category: "breakglass",
-			Value:    "i-have-reason",
-		},
+		return nil, fmt.Errorf("failed to sign breakglass token: %w", err)
```
yolocs, line 184: Copy/paste error? 
```suggestion
		return nil, fmt.Errorf("failed to sign token: %w", err)
```

pkg/cli/token_test.go
```diff
@@ -15,190 +15,176 @@
...
-		t.Errorf("expected length %d to be %d: %#v", len(sigs), want, sigs)
-	} else {
-		headers := sigs[0].ProtectedHeaders()
-		if got, want := headers.Type(), "JWT"; got != want {
-			t.Errorf("typ: expected %q to be %q", got, want)
-		}
-		if got, want := string(headers.Algorithm()), "HS256"; got != want {
-			t.Errorf("alg: expected %q to be %q", got, want)
-		}
+func (j *fakeJVS) CreateJustification(_ context.Context, req *jvspb.CreateJustificationRequest) (*jvspb.CreateJustificationResponse, error) {
```
yolocs, line 159: This is beyond my bar :) Since we're returning whatever token the server returns as a string, I would not really generate a real JTW.

pkg/config/cli_config.go
yolocs: Ack

pkg/config/cli_config.go
yolocs: I'm not a fan of a flat config and I was projecting a more complicated authentication config. We're using the application default credential to generate id token. Cloud Run accepts that. But not with IAP (we can forget about it now) or other authentication middleware (what if it requires an access token instead of id token). That's why I had the sub-struct. I could be overthinking.

yolocs (approved):

pkg/cli/root.go
sethvargo: Nope. I found it returns `ConfigFileNotFoundError` when given a _file_ path that does not exist, but returns `ErrNotExist` when given a _directory_ path that does not exist. `ConfigFileNotFoundError` only seemed to throw when the file had an extension.

pkg/cli/root_test.go
sethvargo: Updated.

pkg/cli/token_test.go
sethvargo: You caught me 😄

pkg/cli/token_test.go
sethvargo: The reason I did this is so we can use the same table-driven tests for testing breakglass tokens and testing tokens. Otherwise the testing logic becomes branching and that's less than ideal.

pkg/config/cli_config.go
sethvargo: The reason I went with the flag config is because it's trivial to accidentally NPE if we're not careful. 

```go
var cfg config.CLIConfig

// later

if cfg.Authentication.Insecure { // npe
}
```

Since neither `Validate` nor `SetDefault` build the deeply-nested struct, it feels risky.

yolocs: It was initialized in SetDefault. But I think it's probably better to keep it simple now.

yolocs (approved):

Issue #135, opened 2022-09-30
Create general-purpose functions in jvspb for printing a JWT as a yaml/json


sqin2019:
follow up of #112

sqin2019:
Not needed anymore, closing.

PR #138, opened 2022-09-30
feat: create jvs token toJson and toYaml funcs

sqin2019:
address #135

### DIFF
--- a/apis/v0/jwt.go
+++ b/apis/v0/jwt.go
@@ -34,7 +34,7 @@ const (
 )
 
 // marshalFn takes an input and return a byte array representation of the input.
-type marshalFn func(in interface{}) ([]byte, error)
+type marshalFn func(v any) ([]byte, error)
 
 // WithTypedJustifications is an option for parsing JWTs that will convert
 // decode the [Justification] claims into the correct Go structure. If this is
@@ -111,12 +111,12 @@ func ClearJustifications(t jwt.Token) error {
 
 // ToJSON converts the token into json with justification claims seperated from other claims.
 func ToJSON(ctx context.Context, t jwt.Token) ([]byte, error) {
-	return marshal(ctx, t, "\n---", func(in interface{}) ([]byte, error) { return json.MarshalIndent(in, "", "  ") })
+	return marshal(ctx, t, "\n---", func(v any) ([]byte, error) { return json.MarshalIndent(v, "", "  ") })
 }
 
 // ToYAML converts the token into yaml with justification claims seperated from other claims.
 func ToYAML(ctx context.Context, t jwt.Token) ([]byte, error) {
-	return marshal(ctx, t, "---", func(in interface{}) ([]byte, error) { return yaml.Marshal(in) })
+	return marshal(ctx, t, "---", func(v any) ([]byte, error) { return yaml.Marshal(v) })
 }
 
 // marshal converts the token into byte array using the given malshal function
--- a/apis/v0/jwt.go
+++ b/apis/v0/jwt.go
@@ -15,10 +15,13 @@
 package v0
 
 import (
+	"context"
+	"encoding/json"
 	"fmt"
 
 	"github.com/lestrrat-go/jwx/v2/jwt"
 	"github.com/mitchellh/mapstructure"
+	"gopkg.in/yaml.v2"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


apis/v0/jwt.go
```diff
@@ -30,6 +33,9 @@ const (
 	jwtJustificationsKey string = "justs"
 )
 
+// marshalFn takes an input and return a byte array representation of the input.
+type marshalFn func(in interface{}) ([]byte, error)
```
capri-xiyue, line 37: nits: use `any` instead of `interface{}`
Since 1.18, Go uses any as the alias for interface{}.

apis/v0/jwt.go
```diff
@@ -102,3 +108,45 @@ func ClearJustifications(t jwt.Token) error {
 
 	return t.Remove(jwtJustificationsKey)
 }
+
+// ToJSON converts the token into json with justification claims seperated from other claims.
+func ToJSON(ctx context.Context, t jwt.Token) ([]byte, error) {
+	return marshal(ctx, t, "\n---", func(v any) ([]byte, error) { return json.MarshalIndent(v, "", "  ") })
```
sethvargo, line 114: This is using MarshalIndent, so it's actually "pretty json". It also doesn't look like we return valid JSON, which could be confusing to callers. L149 appends `"\njustifications\n"` which would make the JSON and YAML invalid to parse.

apis/v0/jwt.go
```diff
@@ -102,3 +108,45 @@ func ClearJustifications(t jwt.Token) error {
...
+func marshal(ctx context.Context, t jwt.Token, sep string, fn marshalFn) ([]byte, error) {
+	if t == nil {
+		return nil, fmt.Errorf("token cannot be nil")
+	}
+
+	claimsMap, err := t.AsMap(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert token into map: %w", err)
+	}
+	delete(claimsMap, jwtJustificationsKey)
```
sethvargo, line 133: Comment why we're deleting this key please

apis/v0/jwt.go
sqin2019: Ah..ok, I can only think of two ways: 1. convert the whole claimsMap to json without separating justs. 2. convert the whole claimsMap to json + create a different method to convert justs to json, so for validate cli command, it will call both funcs it will output justs twice one in the whole json + one separated. Do you have other suggestions?

sqin2019: Is it acceptable to combine claims -> claimsMapNoJusts and justifications -> justs list to a new map, and marshal the new map to json/yaml?

sethvargo: iirc, the original goal here was to pretty-print a JWT on a terminal. If we just want a jwt as JSON, we can call `json.Marshal` directly on the JWT and get _basically_ what you have here.

apis/v0/jwt.go
```diff
@@ -102,3 +108,45 @@ func ClearJustifications(t jwt.Token) error {
...
+	}
+
+	justs, err := GetJustifications(t)
+	if err != nil {
+		return nil, fmt.Errorf("failed to extract justifications: %w", err)
+	}
+	justsOut, err := fn(justs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal justifications as json: %w", err)
+	}
```
yolocs, line 146: Marshal the justifications will end up some unnecessary info:

- marshal result `category: explanation, value: i need access` vs.
- better human readable output `explanation: i need access`

apis/v0/jwt_test.go
```diff
@@ -271,6 +273,138 @@ func TestClearJustifications(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			output, err := ToYAML(context.Background(), tc.token)
+
+			if diff := testutil.DiffErrString(err, tc.expErr); diff != "" {
+				t.Errorf("unexpected err: %s", diff)
+			}
+
+			if !reflect.DeepEqual(output, tc.exp) {
```
yolocs, line 401: For strings (you can convert bytes to string with `string(b)`), please compare with `==`

apis/v0/jwt_test.go
```diff
@@ -271,6 +273,138 @@ func TestClearJustifications(t *testing.T) {
...
+		expErr string
+	}{
+		{
+			name:   "nil_token",
+			token:  nil,
+			expErr: "token cannot be nil",
+		},
+		{
+			name:  "empty_justs",
+			token: testTokenBuilder(t, jwt.NewBuilder().JwtID("test_id")),
```
yolocs, line 361: Could you add a couple of more standard claims, such as `aud` `sub` so we can better view how the output would look like?

yolocs (changes requested):

sqin2019:
Not needed, closing this one as discussed offline with Seth and Chen.

PR #139, opened 2022-10-06
feat: add validate cmd

sqin2019:
address #112
1. make JWKSEndpoint default to server domain if not explicitly specified. 
2. validate cmd output will be in a table with three sections: result, justs, other claims. json and yaml are not considered here since JWT supports both already.

### DIFF
--- a/docs/cli-tool.md
+++ b/docs/cli-tool.md
@@ -36,7 +36,7 @@ following block in the config:
 insecure: true
 ```
 
-JWKS endpoint is also required if you want to validate justification tokens, it will default to the server domain if it is not specified.
+JWKS endpoint is also required if you want to validate justification tokens, it will default to the server domain + `/.well-known/jwks` if it is not specified.
 
 ```yaml
 jwks_endpoint: https://example.com/.well-known/jwks
--- a/docs/cli-tool.md
+++ b/docs/cli-tool.md
@@ -36,8 +36,13 @@ following block in the config:
 insecure: true
 ```
 
-Alternatively, both of these values could be provided via CLI flags `--server`
-and `--insecure`.
+JWKS endpoint is also required if you want to validate justification tokens via CLI, it will default to the server domain if it is not specified.
+
+```yaml
+jwks_endpoint: https://example.com/.well-known/jwks
+```
+
+Alternatively, all of these values could be provided via CLI flags `--server`, `--insecure`, and `--jwks_endpoint`.
 
 ## Command
 
@@ -66,3 +71,13 @@ jvsctl token --breakglass --explanation "jvs is down" --ttl 30m
 In certain cases, we might need to bypass JVS for minting signed JVS tokens.
 E.g. JVS couldn't verify a justification because the ticket system is down. In
 such cases, we can mint break-glass token instead.
+
+We are also able to validate JVS tokens, examples are provided as below.
+
+```shell
+jvsctl validate --token "example token"
+
+# or pass token via pipe
+echo "${JVS_TOKEN}" | jvsctl validate --token -
+cat /tmp/jvs_token | jvsctl validate --token -
+```
--- a/docs/cli-tool.md
+++ b/docs/cli-tool.md
@@ -36,7 +36,7 @@ following block in the config:
 insecure: true
 ```
 
-JWKS endpoint is also required if you want to validate justification tokens via CLI, it will default to the server domain if it is not specified.
+JWKS endpoint is also required if you want to validate justification tokens, it will default to the server domain if it is not specified.
 
 ```yaml
 jwks_endpoint: https://example.com/.well-known/jwks
@@ -78,6 +78,6 @@ We are also able to validate JVS tokens, examples are provided as below.
 jvsctl validate --token "example token"
 
 # or pass token via pipe
-echo "${JVS_TOKEN}" | jvsctl validate --token -
-cat /tmp/jvs_token | jvsctl validate --token -
+echo $JVS_TOKEN | jvsctl validate --token -
+cat /tmp/jvs_token.txt | jvsctl validate --token -
 ```
--- a/pkg/cli/root_test.go
+++ b/pkg/cli/root_test.go
@@ -34,6 +34,7 @@ func TestNewRootCmd(t *testing.T) {
 	if err := os.WriteFile(configFile, []byte(strings.TrimSpace(`
 server: 1.2.3.4:5678
 insecure: true
+jwks_endpoint: "https://1.2.3.4:8080/.well-known/jwks"
 	`)), 0o600); err != nil {
 		t.Fatal(err)
 	}
@@ -44,9 +45,10 @@ insecure: true
 	})
 
 	cases := []struct {
-		name      string
-		args      []string
-		expConfig *config.CLIConfig
+		name            string
+		args            []string
+		expConfig       *config.CLIConfig
+		expJWKSEndpoint string
 	}{
 		{
 			name: "default_config",
@@ -54,15 +56,18 @@ insecure: true
 				Version: "1",
 				Server:  "127.0.0.1:8080",
 			},
+			expJWKSEndpoint: "https://127.0.0.1:8080/.well-known/jwks",
 		},
 		{
 			name: "flag_config",
 			args: []string{"--config", configFile},
 			expConfig: &config.CLIConfig{
-				Version:  "1",
-				Server:   "1.2.3.4:5678",
-				Insecure: true,
+				Version:      "1",
+				Server:       "1.2.3.4:5678",
+				JWKSEndpoint: "https://1.2.3.4:8080/.well-known/jwks",
+				Insecure:     true,
 			},
+			expJWKSEndpoint: "https://1.2.3.4:8080/.well-known/jwks",
 		},
 		{
 			name: "flag_server",
@@ -71,6 +76,7 @@ insecure: true
 				Version: "1",
 				Server:  "1.2.3.4:5678",
 			},
+			expJWKSEndpoint: "https://1.2.3.4:8080/.well-known/jwks",
 		},
 		{
 			name: "flag_insecure",
@@ -80,6 +86,17 @@ insecure: true
 				Server:   "127.0.0.1:8080",
 				Insecure: true,
 			},
+			expJWKSEndpoint: "https://127.0.0.1:8080/.well-known/jwks",
+		},
+		{
+			name: "flag_jwks_endpoint",
+			args: []string{"--jwks_endpoint", "https://1.2.3.4:8080/.well-known/jwks"},
+			expConfig: &config.CLIConfig{
+				Version:      "1",
+				Server:       "127.0.0.1:8080",
+				JWKSEndpoint: "https://1.2.3.4:8080/.well-known/jwks",
+			},
+			expJWKSEndpoint: "https://1.2.3.4:8080/.well-known/jwks",
 		},
 	}
 
@@ -106,6 +123,9 @@ insecure: true
 
 			if diff := cmp.Diff(tc.expConfig, &cfg); diff != "" {
 				t.Errorf("config (-want, +got):\n%s", diff)
+			}
+			if diff := cmp.Diff(tc.expJWKSEndpoint, cfg.GetJWKSEndpoint()); diff != "" {
+				t.Errorf("jwksEndpoint (-want, +got):\n%s", diff)
 			}
 		})
 	}
--- a/pkg/cli/root_test.go
+++ b/pkg/cli/root_test.go
@@ -45,18 +45,17 @@ jwks_endpoint: "https://1.2.3.4:8080/.well-known/jwks"
 	})
 
 	cases := []struct {
-		name            string
-		args            []string
-		expConfig       *config.CLIConfig
-		expJWKSEndpoint string
+		name      string
+		args      []string
+		expConfig *config.CLIConfig
 	}{
 		{
 			name: "default_config",
 			expConfig: &config.CLIConfig{
-				Version: "1",
-				Server:  "127.0.0.1:8080",
+				Version:      "1",
+				Server:       "127.0.0.1:8080",
+				JWKSEndpoint: "https://127.0.0.1:8080/.well-known/jwks",
 			},
-			expJWKSEndpoint: "https://127.0.0.1:8080/.well-known/jwks",
 		},
 		{
 			name: "flag_config",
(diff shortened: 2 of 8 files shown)


myurtoglu:
Added @yolocs for Go readability.

docs/cli-tool.md
```diff
@@ -66,3 +71,13 @@ jvsctl token --breakglass --explanation "jvs is down" --ttl 30m
...
 E.g. JVS couldn't verify a justification because the ticket system is down. In
 such cases, we can mint break-glass token instead.
+
+We are also able to validate JVS tokens, examples are provided as below.
+
+```shell
+jvsctl validate --token "example token"
+
+# or pass token via pipe
+echo "${JVS_TOKEN}" | jvsctl validate --token -
```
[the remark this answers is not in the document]
sqin2019: both works, will go for the easier one, thanks!

pkg/cli/validate.go
```diff
@@ -0,0 +1,191 @@
...
+			opts.token = string(buf)
+		} else {
+			// Use user input
+			fmt.Print("Enter token: ")
+			fmt.Scanf("%s", &opts.token)
+		}
+	}
+
+	// Validate token
+	breakglass := false
```
yolocs, line 127: Can you break the following logic to be 3 functions? 1. print breakglass 2. print justifications 3. print standard claims

pkg/config/cli_config.go
```diff
@@ -55,3 +60,11 @@ func (cfg *CLIConfig) SetDefault() {
...
 	}
 }
+
+// Default to server's doman if JWKSEndpoint is not specified.
+func (cfg *CLIConfig) GetJWKSEndpoint() string {
+	if cfg.JWKSEndpoint == "" && cfg.Server != "" {
+		return fmt.Sprintf("https://%s:8080/.well-known/jwks", strings.Split(cfg.Server, ":")[0])
+	}
+	return cfg.JWKSEndpoint
+}
```
yolocs, line 70: Can we make this part of `SetDefault()` instead of a separate func?

pkg/cli/validate.go
```diff
@@ -0,0 +1,191 @@
...
+		Args: cobra.ExactArgs(0),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runValidateCmd(cmd, opts, args)
+		},
+	}
+
+	flags := cmd.Flags()
+	flags.StringVarP(&opts.token, "token", "t", "",
+		"The JVS token that needs validation")
+	cmd.MarkFlagRequired("token") //nolint // not expect err
```
yolocs, line 86: Can we link to the issue (or create one) to output in raw json? E.g. `--format=json`.

pkg/cli/validate_test.go
```diff
@@ -0,0 +1,265 @@
...
+				stdout, _, gotErr = testExecuteCommand(t, cmd, tc.args...)
+			}
+			if diff := testutil.DiffErrString(gotErr, tc.expErr); diff != "" {
+				t.Fatal(diff)
+			}
+			if gotErr != nil {
+				return
+			}
+
+			if diff := cmp.Diff(tc.expOut, stdout); diff != "" {
```
yolocs, line 227: Do we capture the error output? Should we compare them?

yolocs (changes requested):

pkg/cli/validate.go
```diff
@@ -0,0 +1,191 @@
...
+    # Validate the justification token string
+    jvsctl validate --token "example token string"
+
+    # Validate the justification token read from pipe
+    cat token.txt | jvsctl token --token -
+
+    # Output
+	-------RESULT-------
+	validated!
+	breakglass  false
```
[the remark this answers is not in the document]
sqin2019: I kept "validated!" to be explicit, also to echo the name of this cmd.

pkg/cli/root_test.go
```diff
@@ -107,6 +124,9 @@ insecure: true
 			if diff := cmp.Diff(tc.expConfig, &cfg); diff != "" {
 				t.Errorf("config (-want, +got):\n%s", diff)
 			}
+			if diff := cmp.Diff(tc.expJWKSEndpoint, cfg.GetJWKSEndpoint()); diff != "" {
+				t.Errorf("jwksEndpoint (-want, +got):\n%s", diff)
+			}
```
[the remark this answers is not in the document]
sethvargo: +1 - I'm not sure what we need this in the struct; we want to make sure it's in the config.

pkg/cli/validate.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/abcxyz/jvs/client-lib/go/client"
+	"github.com/abcxyz/jvs/pkg/config"
+	"github.com/spf13/cobra"
+)
+
+// validateCmdOptions holds all the inputs and flags for the validate subcommand.
+type validateCmdOptions struct {
+	config *config.CLIConfig
+
+	token string
```
sethvargo, line 38: Nit: doc

pkg/cli/validate.go
```diff
@@ -0,0 +1,191 @@
...
+		JWKSEndpoint:    opts.config.GetJWKSEndpoint(),
+		CacheTimeout:    5 * time.Minute,
+		AllowBreakglass: true,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create jvs client: %w", err)
+	}
+
+	if opts.token == "-" {
+		stdin, ok := cmd.InOrStdin().(*os.File)
```
sethvargo, line 105: Why do we need to require this to be a file? I'm not sure I see why we have to handle the pipe differently.

sethvargo, line 105: I'm probably missing something, but I think we just want to `ioutil.ReadAll(io.LimitReader(stdin, 64*1_000))`. I believe that will read until EOF. It doesn't handle the edge case where the terminal is not a TTY, but that can be a later enhancement.

pkg/cli/validate_test.go
```diff
@@ -0,0 +1,265 @@
...
+		{
+			name: "signed",
+			config: &config.CLIConfig{
+				JWKSEndpoint: svr.URL + path,
+			},
+			args: []string{"-t", signedToken},
+			expOut: `-------RESULT-------
+validated!
+breakglass  false
+----JUSTIFICATION----
```
sethvargo, line 132: Nit, I'd expect a newline between sections

pkg/config/cli_config.go
sethvargo: Agreed. Also naming functions Get/Set is generally frowned upon outside of auto-generated code.

pkg/config/cli_config.go
```diff
@@ -55,3 +60,11 @@ func (cfg *CLIConfig) SetDefault() {
 		cfg.Version = "1"
 	}
 }
+
+// Default to server's doman if JWKSEndpoint is not specified.
+func (cfg *CLIConfig) GetJWKSEndpoint() string {
+	if cfg.JWKSEndpoint == "" && cfg.Server != "" {
+		return fmt.Sprintf("https://%s:8080/.well-known/jwks", strings.Split(cfg.Server, ":")[0])
```
sethvargo, line 67: This can panic

pkg/cli/validate_test.go
```diff
@@ -0,0 +1,265 @@
...
+	for _, tc := range cases {
+		tc := tc
+
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			cmd := newValidateCmd(tc.config)
+			var stdout string
+			var gotErr error
+			if tc.pipe {
```
sethvargo, line 197: I think we're over-complicating things here.

sqin2019: Any suggestions?

pkg/cli/validate_test.go
sqin2019: Is it necessary? the error output is the token flag usage.

pkg/cli/validate.go
sqin2019: The usage of it is added when it is added as a flag, will the doc here be redundant?

pkg/config/cli_config.go
sqin2019: Could you explain?

sethvargo: Sorry - I missed the check above.

pkg/cli/validate.go
sethvargo: The usage is for the CLI user. The docs are for future developers.

pkg/cli/validate.go
sqin2019: Created 2. & 3. helper functions, keeping 1 as it is since it is simple enough.

pkg/cli/validate.go
sqin2019: updated.

pkg/cli/validate.go
sqin2019: Updated, could you check again? thanks!

pkg/cli/validate.go
```diff
@@ -0,0 +1,194 @@
...
+	for _, j := range justs {
+		if _, err := fmt.Fprintf(w, "%s\t\"%s\"\n", j.GetCategory(), j.GetValue()); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func writeStandardClaims(ctx context.Context, w *tabwriter.Writer, token jwt.Token) error {
+	// Convert standard claims into a map, excluding justifications claim which is already handled
```
yolocs, line 165:
```suggestion
	// Convert standard claims into a map, excluding justifications claim which is printed separately
```
nit

pkg/cli/validate.go
```diff
@@ -0,0 +1,194 @@
...
+		return fmt.Errorf("failed to convert token into map: %w", err)
+	}
+	delete(claimsMap, "justs")
+	claims := make(map[string]string, len(claimsMap))
+	claimsKeys := make([]string, 0, len(claimsMap))
+	for k, v := range claimsMap {
+		claimsKeys = append(claimsKeys, k)
+		jv, err := json.Marshal(v)
+		if err != nil {
+			claims[k] = fmt.Sprintf("failed to marshal to json: %s", err)
```
yolocs, line 177: Why don't we return the error?

pkg/cli/validate.go
```diff
@@ -0,0 +1,194 @@
...
+	if _, err := fmt.Fprintf(w, "-----BREAKGLASS-----\n%t\n", breakglass); err != nil {
+		return err
+	}
+
+	if err := writeJustification(w, token); err != nil {
+		return err
+	}
+	if err := writeStandardClaims(ctx, w, token); err != nil {
+		return err
+	}
```
yolocs, line 143: For these errors, I recommend wrapping them like:

```
return fmt.Errorf("justification token is valid but failed to print justifications: %w", err)
```

pkg/cli/validate_test.go
yolocs: For the `io.Reader`, you can use https://pkg.go.dev/strings#Builder instead of writing to a file

pkg/cli/validate_test.go
yolocs: You're right. `DiffErrString` already does that.

pkg/cli/validate_test.go
```diff
@@ -0,0 +1,268 @@
...
+	token, err := jwt.NewBuilder().
+		Audience([]string{justification.DefaultAudience}).
+		IssuedAt(now).
+		Issuer(Issuer).
+		JwtID("test-jwt").
+		NotBefore(now).
+		Subject(Subject).
+		Build()
+	if err != nil {
+		tb.Fatalf("failed to create unsigned token: %s", err)
```
yolocs, line 250:
```suggestion
		tb.Fatalf("failed to create unsigned token: %v", err)
```

pkg/cli/validate_test.go
```diff
@@ -0,0 +1,268 @@
...
+	tb.Helper()
+
+	headers := jws.NewHeaders()
+	if err := headers.Set(jws.KeyIDKey, keyID); err != nil {
+		tb.Fatal(err)
+	}
+
+	token, err := jwt.Sign(unsignedToken, jwt.WithKey(jwa.ES256, privateKey, jws.WithProtectedHeaders(headers)))
+	if err != nil {
+		tb.Fatalf("failed to sign token: %s", err)
```
yolocs, line 265:
```suggestion
		tb.Fatalf("failed to sign token: %v", err)
```

yolocs (changes requested):

pkg/cli/validate.go
```diff
@@ -0,0 +1,194 @@
...
+    jvsctl validate --token "example token string"
+
+    # Validate the justification token read from pipe
+    cat token.txt | jvsctl validate --token -
+
+    # Output
+    ------BREAKGLASS------
+    false
+
+    ----JUSTIFICATIONs----
```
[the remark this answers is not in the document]
sqin2019: typo...

pkg/cli/validate.go
sqin2019: Yeah, we could, to be consistent, updating to return error.

pkg/cli/validate_test.go
sqin2019: accidently found out that strings.NewReader(signedToken) is working as well?

yolocs (approved):

yolocs:
Please fix the go_lint issue.

docs/cli-tool.md
```diff
@@ -36,8 +36,13 @@ following block in the config:
 insecure: true
 ```
 
-Alternatively, both of these values could be provided via CLI flags `--server`
-and `--insecure`.
+JWKS endpoint is also required if you want to validate justification tokens, it will default to the server domain if it is not specified.
```
sethvargo, line 39:
```suggestion
JWKS endpoint is also required if you want to validate justification tokens, it will default to the server domain + `/.well-known/jwks` if it is not specified.
```

pkg/cli/validate.go
```diff
@@ -0,0 +1,193 @@
...
+	return cmd
+}
+
+func runValidateCmd(cmd *cobra.Command, opts *validateCmdOptions, args []string) error {
+	ctx := context.Background()
+
+	jvsclient, err := client.NewJVSClient(ctx, &client.JVSConfig{
+		Version:         "1",
+		JWKSEndpoint:    opts.config.JWKSEndpoint,
+		CacheTimeout:    5 * time.Minute,
```
sethvargo, line 99: Two things here:

1. This should be a constant.
2. Does caching actually make sense? This isn't a long-running service, so the cache is effectively expired once the CLI exits since it's all in-memory.

pkg/cli/validate_test.go
```diff
@@ -0,0 +1,250 @@
...
+	breakglassToken, err := jvspb.CreateBreakglassToken(token, "prod is down")
+	if err != nil {
+		t.Fatalf("failed to build breakglass token: %v", err)
+	}
+
+	cases := []struct {
+		name   string
+		config *config.CLIConfig
+		args   []string
+		pipe   bool
```
sethvargo, line 101: Nit: `pipe` is generic and not really descriptive. Maybe `writeTokenToStdin` or have this as `stdin` and accept io.Reader, which drops the branching logic in the test.

```go
cases := []struct{}{
  // ...
  stdin io.Reader
}{
  {
    name: "from_stdin",
    // ...
    stdin: strings.NewReader(signedToken)
  },
}

for _, tc := range cases {
  tc := tc

  t.Run(tc.name, func(t *testing.T) {
    // ...

    stdout, stderr, gotErr = testExecuteCommandStdin(t, cmd, tc,stdin, tc.args...)
  })
}
```

Since `stdin` will default to `nil`, everything works.

pkg/cli/validate_test.go
```diff
@@ -0,0 +1,250 @@
...
+			}
+
+			if diff := cmp.Diff(tc.expOut, stdout); diff != "" {
+				t.Errorf("output: diff (-want, +got):\n%s", diff)
+			}
+		})
+	}
+}
+
+func testTokenBuilder(tb testing.TB) jwt.Token {
```
sethvargo, line 219: Can you file a follow-up bug - we should probably expose this as a top-level helper in the api package.

sethvargo (approved):

pkg/cli/validate_test.go
sqin2019: #142 created.

pkg/cli/validate.go
sqin2019: discussed offline, jvs config requires a >0 cache timeout, made it a constant and added comment.

PR #141, opened 2022-10-10
Parallelize public key downloads

### LINKED ISSUE #34: Consider parallelizing public key lookups.
we look up the keys in the "jwkList" method of key_hosting n an iterative fashion. We could do this in parallel in order to speed it up.

sethvargo:
Fixes #34


pkg/jvscrypto/kmsutil.go
```diff
@@ -93,38 +94,108 @@ func getLabelValue(versionName string) (string, error) {
...
 
-		publicKeyResp, err := client.GetPublicKey(ctx, &kmspb.GetPublicKeyRequest{
-			Name: keyVersion.Name,
-		})
-		if err != nil {
-			return nil, fmt.Errorf("failed to get public key for key version %s: %w", keyVersion.Name, err)
+		for _, v := range result.Value {
+			keyVersions[v] = struct{}{}
 		}
+	}
```
yolocs, line 142: This logic is to deduplicate key versions. But I wonder why it's needed?

yolocs (approved):

pkg/jvscrypto/kmsutil.go
sethvargo: Since `parentKeys` is a slice, there's nothing that prevents a user from giving duplicate parent keys names. That would result in us returning multiple key versions.

It would be the same if we de-duped parentKeys, but this felt more defensive.

sqin2019 (approved):

Issue #129, opened 2022-09-26
Set up initial release


yolocs:

sethvargo:
Similar to https://github.com/abcxyz/lumberjack/issues/286, I think we should consider looking at goreleaser here to automate some of this. It has native support for GitHub Actions and integrates with cosign and stuff.

yolocs:
Should be done. Pending manual testing

PR #145, opened 2022-10-18
chore: set up initial release workflow

yolocs:
Part of https://github.com/abcxyz/jvs/issues/129
Missing maven release at the moment.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,7 @@ env:
   WIF_PROVIDER: 'projects/1096923323432/locations/global/workloadIdentityPools/github-pool/providers/github-provider'
   WIF_SERVICE_ACCOUNT: 'gh-access-sa@jvs-ci-test.iam.gserviceaccount.com'
 
+# Don't cancel in progress since we don't want to have half-baked release.
 concurrency: '${{ github.workflow }}-${{ github.head_ref || github.ref }}'
 
 jobs:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,9 +9,7 @@ env:
   WIF_PROVIDER: 'projects/1096923323432/locations/global/workloadIdentityPools/github-pool/providers/github-provider'
   WIF_SERVICE_ACCOUNT: 'gh-access-sa@jvs-ci-test.iam.gserviceaccount.com'
 
-concurrency:
-  group: '${{ github.workflow }}-${{ github.head_ref || github.ref }}'
-  cancel-in-progress: true
+concurrency: '${{ github.workflow }}-${{ github.head_ref || github.ref }}'
 
 jobs:
   unit-test:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+      - 'v*'
+env:
+  # .goreleaser.docker.yaml reads CONTAINER_REGISTRY
+  CONTAINER_REGISTRY: 'us-docker.pkg.dev/abcxyz-artifacts/docker-images'
+  WIF_PROVIDER: 'projects/1096923323432/locations/global/workloadIdentityPools/github-pool/providers/github-provider'
+  WIF_SERVICE_ACCOUNT: 'gh-access-sa@jvs-ci-test.iam.gserviceaccount.com'
+
+concurrency:
+  group: '${{ github.workflow }}-${{ github.head_ref || github.ref }}'
+  cancel-in-progress: true
```
sethvargo, line 14: I'm not sure we want to cancel in-progress releases? Seems bad to have something half-released.

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+          token_format: 'access_token'
+      - name: 'Authenticate to Artifact Registry'
+        uses: 'docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a' # ratchet:docker/login-action@v2
+        with:
+          username: 'oauth2accesstoken'
+          password: '${{ steps.auth.outputs.access_token }}'
+          registry: 'us-docker.pkg.dev'
+      - uses: 'goreleaser/goreleaser-action@b508e2e3ef3b19d4e4146d4f8fb3ba9db644a757' # ratchet:goreleaser/goreleaser-action@v3
+        with:
+          version: 'latest'
```
sethvargo, line 52: We should pin

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+          args: 'release -f .goreleaser.docker.yaml --rm-dist'
+        env:
```
sethvargo, line 54: If the env is defined above, I don't think you need to define it again here. I think all jobs inherit it.

.github/workflows/release.yml
yolocs: Manually pinned to a version

.github/workflows/release.yml
```diff
@@ -9,9 +9,7 @@ env:
   WIF_PROVIDER: 'projects/1096923323432/locations/global/workloadIdentityPools/github-pool/providers/github-provider'
   WIF_SERVICE_ACCOUNT: 'gh-access-sa@jvs-ci-test.iam.gserviceaccount.com'
 
-concurrency:
-  group: '${{ github.workflow }}-${{ github.head_ref || github.ref }}'
-  cancel-in-progress: true
+concurrency: '${{ github.workflow }}-${{ github.head_ref || github.ref }}'
```
sethvargo, line 12: Also add a comment about why we don't cancel-in-progress please.

sethvargo (approved):

PR #146, opened 2022-10-18
chore: prepare pom.xml for maven package release

yolocs:
Part of https://github.com/abcxyz/jvs/issues/129

In the release workflow, we'd a step to run the following command to publish the package.

```
# [VERSION] like 0.0.1
mvn clean deploy -Drevision=[VERSION]
```

Ref: https://maven.apache.org/maven-ci-friendly.html

Maven packages are immutable and can't be overwritten. If needed, we would need to delete the existing version before publishing new ones. To delete, we can run:

```
gcloud artifacts versions delete [VERSION] --package=com.abcxyz.jvs:jvs --project=abcxyz-artifacts --repository=maven-repo --location=us --quiet
```

### DIFF
--- a/client-lib/java/pom.xml
+++ b/client-lib/java/pom.xml
@@ -5,7 +5,7 @@
   <groupId>com.abcxyz.jvs</groupId>
   <artifactId>jvs-client</artifactId>
   <packaging>jar</packaging>
-  <version>0.0.1-SNAPSHOT</version>
+  <version>${revision}</version>
 
   <dependencyManagement>
     <dependencies>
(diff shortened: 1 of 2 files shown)


client-lib/java/pom.xml
```diff
@@ -5,7 +5,7 @@
   <groupId>com.abcxyz.jvs</groupId>
   <artifactId>jvs-client</artifactId>
   <packaging>jar</packaging>
-  <version>0.0.1-SNAPSHOT</version>
+  <version>${revision}</version>
```
raserva, line 8: is this the git revision?

yolocs: Right. We will create git tags like `v0.0.1` and the release version will be that without the `v`.
For ref: https://maven.apache.org/maven-ci-friendly.html

raserva: That link seems to recommend the opposite - that it should be specific in the pom file. 

```
But wait there is a problem? Which version will the artifacts have? So you need to define the version for your artifacts. The first possibility is to use the command line like this:

mvn -Drevision=1.0.0-SNAPSHOT clean package

This will become cumbersome over the time. So the other solution for this is to simply use a property inside the pom file which looks like this:


<properties>
    <revision>1.0.0-SNAPSHOT</revision>
  </properties>
```

yolocs: IIUC, the problem is about _not_ having `${revision}` being a property with a default value. But I've added it as a property.

raserva (approved):

client-lib/java/pom.xml
sethvargo: Can we pull the VCS information from git instead of requiring a user to supply it?

yolocs: I didn't find a native mvn way to auto pull in VCS info. My idea is to run `mvn clean deploy -Drevision=[VERSION]` in the release workflow where we would pass the git tag to `-Drevision`.

PR #155, opened 2022-10-28
Add internal/version and print version on boot

sethvargo:


yolocs (approved):
👍

internal/version/version.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+var (
+	// Name is the name of the binary. This can be overridden by the build
+	// process.
+	Name = "jvs"
+
+	// Version is the main package version. This can be overridden by the build
+	// process.
+	Version = "source"
```
sqin2019, line 29: Hi, just wondering what are other Version values we are expecting other than "source", and when the Name and Version vars will be overridden?

sethvargo: Hey @sqin2019 - they are overwritten by the build command: https://github.com/abcxyz/jvs/blob/3103e051920d7949d4b692b070032c72f43314de/.goreleaser.yaml#L38-L40

This allows us to inject the git tag and git commit directly into the build.

sqin2019: Got it, thank you!

PR #163, opened 2023-01-07
fix: add a check on new keys before return shouldRotate

### LINKED ISSUE #160: Fix cert rotation shouldRotate bug
code link: https://github.com/abcxyz/jvs/blob/e99deb918b1221e4b0430003fc15a0319ffabd49/pkg/jvscrypto/rotation_handler.go#L160 

Currently implementation will keep generating new key versions when primary key version is older than [rotation age](https://github.com/abcxyz/jvs/blob/39392f6bd7b61c0250eddefc6a7504ba55898e51/pkg/config/crypto_config.go#L84) even there are newer key versions exist. 

This will become an issue when primary key age > rotation age, and rotation service is triggered at a frequency < [propagation](https://github.com/abcxyz/jvs/blob/39392f6bd7b61c0250eddefc6a7504ba55898e51/pkg/jvscrypto/rotation_handler.go#L246) time, because it would keep generating new key versions, and the new key versions would never get promoted, and primary key version stays the same. 

fix: stop rotating key versions when there are newer (than primary) key versions.

sqin2019:
fix #160 and some typo.

1. Adding a check on primary == nil in shouldPromote will promote the newest key if primary is nil regardless of propagation delay. 
2. Adding a check on newest == nil in shouldRotate will stop generating new key if new key already exists regardless of primary key's rotation age.

### DIFF
--- a/pkg/jvscrypto/rotation_handler.go
+++ b/pkg/jvscrypto/rotation_handler.go
@@ -226,7 +226,9 @@ func (h *RotationHandler) shouldDisable(ctx context.Context, ver *kmspb.CryptoKe
 	return shouldDisable
 }
 
-func (h *RotationHandler) shouldRotate(ctx context.Context, primary *kmspb.CryptoKeyVersion, newest *kmspb.CryptoKeyVersion, curTime time.Time) bool {
+// Determines whether a new key version should be created. It returns true only when the newest version does not exist and the
+// primary version reaches its rotation age.
+func (h *RotationHandler) shouldRotate(ctx context.Context, primary, newest *kmspb.CryptoKeyVersion, curTime time.Time) bool {
 	logger := logging.FromContext(ctx)
 	if newest != nil {
 		logger.Debug("new version already created, no action necessary.", zap.Any("version", newest))
@@ -242,7 +244,9 @@ func (h *RotationHandler) shouldRotate(ctx context.Context, primary *kmspb.Crypt
 	return shouldRotate
 }
 
-func (h *RotationHandler) shouldPromote(ctx context.Context, primary *kmspb.CryptoKeyVersion, newest *kmspb.CryptoKeyVersion, curTime time.Time) bool {
+// Determines whether the newest key version should be promoted to primary. It returns true when the primary verison does not exist
+// or when the newest version crosses its propagation delay.
+func (h *RotationHandler) shouldPromote(ctx context.Context, primary, newest *kmspb.CryptoKeyVersion, curTime time.Time) bool {
 	logger := logging.FromContext(ctx)
 	if newest == nil {
 		return false
--- a/pkg/jvscrypto/rotation_handler.go
+++ b/pkg/jvscrypto/rotation_handler.go
@@ -228,7 +228,7 @@ func (h *RotationHandler) shouldDisable(ctx context.Context, ver *kmspb.CryptoKe
 
 func (h *RotationHandler) shouldRotate(ctx context.Context, primary *kmspb.CryptoKeyVersion, newest *kmspb.CryptoKeyVersion, curTime time.Time) bool {
 	logger := logging.FromContext(ctx)
-	if (newest != nil) {
+	if newest != nil {
 		logger.Debug("new version already created, no action necessary.", zap.Any("version", newest))
 		return false
 	}
--- a/pkg/jvscrypto/rotation_handler.go
+++ b/pkg/jvscrypto/rotation_handler.go
@@ -145,7 +145,7 @@ func (h *RotationHandler) actionsForNewVersions(ctx context.Context, vers []*kms
 	newest := newestEnabledVer(vers)
 
 	// If newest is eligible for promotion, promote and don't do anything with the current primary.
-	if h.shouldPromote(ctx, newest, curTime) {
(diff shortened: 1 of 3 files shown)


pkg/jvscrypto/rotation_handler.go
```diff
@@ -221,34 +221,42 @@ func (h *RotationHandler) shouldDisable(ctx context.Context, ver *kmspb.CryptoKe
...
 		logger.Info("version created before cutoff date, should disable.", zap.Any("version", ver), zap.Any("cutoff", cutoff))
 	} else {
-		logger.Debug("version created before cutoff date, no action necessary.", zap.Any("version", ver), zap.Any("cutoff", cutoff))
+		logger.Debug("version created after cutoff date, no action necessary.", zap.Any("version", ver), zap.Any("cutoff", cutoff))
 	}
 	return shouldDisable
 }
 
-func (h *RotationHandler) shouldRotate(ctx context.Context, ver *kmspb.CryptoKeyVersion, curTime time.Time) bool {
+func (h *RotationHandler) shouldRotate(ctx context.Context, primary *kmspb.CryptoKeyVersion, newest *kmspb.CryptoKeyVersion, curTime time.Time) bool {
```
sethvargo, line 229:
```suggestion
func (h *RotationHandler) shouldRotate(ctx context.Context, primary, newest *kmspb.CryptoKeyVersion, curTime time.Time) bool {
```

pkg/jvscrypto/rotation_handler.go
```diff
@@ -221,34 +221,42 @@ func (h *RotationHandler) shouldDisable(ctx context.Context, ver *kmspb.CryptoKe
...
+		logger.Info("version created before cutoff date, should rotate.", zap.Any("version", primary), zap.Any("cutoff", cutoff))
 	} else {
-		logger.Debug("version created before cutoff date, no action necessary.", zap.Any("version", ver), zap.Any("cutoff", cutoff))
+		logger.Debug("version created after cutoff date, no action necessary.", zap.Any("version", primary), zap.Any("cutoff", cutoff))
 	}
 	return shouldRotate
 }
 
-func (h *RotationHandler) shouldPromote(ctx context.Context, ver *kmspb.CryptoKeyVersion, curTime time.Time) bool {
+func (h *RotationHandler) shouldPromote(ctx context.Context, primary *kmspb.CryptoKeyVersion, newest *kmspb.CryptoKeyVersion, curTime time.Time) bool {
```
sethvargo, line 245:
```suggestion
func (h *RotationHandler) shouldPromote(ctx context.Context, primary, newest *kmspb.CryptoKeyVersion, curTime time.Time) bool {
```

pkg/jvscrypto/rotation_handler.go
```diff
@@ -221,34 +221,42 @@ func (h *RotationHandler) shouldDisable(ctx context.Context, ver *kmspb.CryptoKe
 	if shouldDisable {
 		logger.Info("version created before cutoff date, should disable.", zap.Any("version", ver), zap.Any("cutoff", cutoff))
 	} else {
-		logger.Debug("version created before cutoff date, no action necessary.", zap.Any("version", ver), zap.Any("cutoff", cutoff))
+		logger.Debug("version created after cutoff date, no action necessary.", zap.Any("version", ver), zap.Any("cutoff", cutoff))
```
sethvargo, line 224: It can be a followup, but we can use `Debugw` instead of `Debug` to clean this up:

```suggestion
		logger.Debugw("version created after cutoff date, no action necessary.", "version", "cutoff", cutoff)
```

pkg/jvscrypto/rotation_handler.go
sethvargo, line 229: Can we add some documentation to this now about the conditions under which it returns `true` and the conditions under which it returns `false`?

sethvargo (approved):

pkg/jvscrypto/rotation_handler.go
sqin2019: Sure, will do this as a follow up. Do you mean we prefer Debugw than Debug and Debugf in general? If so, saw a few other places using Debug and Debugf, can clean them up together.

pkg/jvscrypto/rotation_handler.go
sqin2019: Updated.

pkg/jvscrypto/rotation_handler.go
sethvargo: I prefer Debugw because we're not really concerned with allocations.

capri-xiyue (approved):

pkg/jvscrypto/rotation_handler_test.go
```diff
@@ -162,6 +172,15 @@ func TestDetermineActions(t *testing.T) {
...
 		},
+		{
+			name: "single_key_new",
+			versions: []*kmspb.CryptoKeyVersion{
+				newEnabledKey2,
+			},
+			wantActions: []*actionTuple{
+				{ActionPromote, newEnabledKey2},
+			},
+		},
```
yolocs, line 183: Do we also want to test multiple new keys?

sqin2019: Sure, will fix it in my next PR.

PR #164, opened 2023-01-10
clean up: replace print style logging with loose-typed structured style

sqin2019:
follow ups of https://github.com/abcxyz/jvs/pull/163 open conversations.

### DIFF
--- a/pkg/jvscrypto/rotation_handler_test.go
+++ b/pkg/jvscrypto/rotation_handler_test.go
@@ -223,6 +223,16 @@ func TestDetermineActions(t *testing.T) {
 			},
 		},
 		{
+			name: "three_enabled_keys_no_action",
+			versions: []*kmspb.CryptoKeyVersion{
+				oldEnabledKey,
+				newEnabledKey,
+				newEnabledKey2,
+			},
+			primary:     oldEnabledKey.Name,
+			wantActions: []*actionTuple{},
+		},
+		{
 			name: "three_enabled_keys",
 			versions: []*kmspb.CryptoKeyVersion{
 				oldEnabledKey,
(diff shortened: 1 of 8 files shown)


pkg/jvscrypto/rotation_handler_test.go
```diff
@@ -222,6 +222,16 @@ func TestDetermineActions(t *testing.T) {
 				{ActionPromote, newEnabledKey},
 			},
 		},
+		{
+			name: "three_enabled_keys_no_action",
+			versions: []*kmspb.CryptoKeyVersion{
+				oldEnabledKey,
+				newEnabledKey,
+				newEnabledKey2,
```
yolocs, line 230: I'm a little confused. When would we promote a new key and which one would we promote (`newEnabledKey` or `newEnabledKey2`)?

sqin2019: We only look at primary and the newest key when determining shouldPromote and shouldRotate. 
1. should promote the newest key to primary when either the primary does not exist or it is > propagation
2. should rotate only when primary is > rotation age && if the newest key does not exist.

For 1, reasons are:
- we don't expect any keys between primary and the newest key, because shouldRotate should prevent that. 
-  It is hard to predict the outcome if we try to check each key between primary and the newest key and determine if they should be promoted.

yolocs: So should we change this test to validate we will always promote the newest key?

sqin2019: ok, I see, this one checks it wont promote any key, I will add an additional one that does promote.

sethvargo (approved):

yolocs (approved):

Issue #140, opened 2022-10-07
Add output format flag (json/yaml) to cli validate cmd and output accordingly.


sqin2019:
follow up of #112

PR #167, opened 2023-01-12
Add formatters for validating the token

sethvargo:
Closes GH-140

### DIFF
--- a/pkg/formatter/text.go
+++ b/pkg/formatter/text.go
@@ -1,0 +1,213 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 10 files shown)


pkg/formatter/text.go
```diff
@@ -0,0 +1,213 @@
...
+
+	jvspb "github.com/abcxyz/jvs/apis/v0"
+	"github.com/lestrrat-go/jwx/v2/jwt"
+)
+
+// Text outputs the token into three sections:
+//   - breakglass
+//   - justification
+//   - standard claims
+type Text struct{}
```
yolocs, line 36: This is fancy :) Maybe worth a `cliutil` in abcxyz/pkg.

yolocs (approved):

PR #168, opened 2023-01-19
Switch to using slog

sethvargo:
@yolocs this is _basically_ what it would look like to migrate from zap->slog. We drop 4 deps and gain 1, so net change is 3. Once slog moves to stdlib, we can drop the dependency on golang.org/x/exp.

It's largely just a find-and-replace for most things except `zap.Errorw`. `slog.Error` actually accepts an "error" whereas `zap.Errorw` was variadic.


bradegler (approved):

capri-xiyue (approved):

sethvargo:
Not gonna merge this into slog is in the stdlib

sethvargo:
I'm gonna close this out until we're closer to 1.21 (around August)

Issue #169, opened 2023-01-24
TTL should have a sane reasonable limit


yolocs:
### TL;DR

In CLI, we default to 15m for token TTL (https://github.com/abcxyz/jvs/blob/main/pkg/cli/token.go#L83). In the server, we simply pass it over to generate the token expiration time https://github.com/abcxyz/jvs/blob/main/pkg/justification/processor.go#L167. It's possible that a user can generate a token that's valid for a very long time, which is not desirable. We should set a reasonable upper limit, e.g. 1d.

PR #172, opened 2023-01-24
fix: Add limit for token TTL upper bound

sailorlqh:
Fix issue: https://github.com/abcxyz/jvs/issues/169

### DIFF
--- a/pkg/justification/processor.go
+++ b/pkg/justification/processor.go
@@ -164,7 +164,11 @@ func (p *Processor) createToken(ctx context.Context, now time.Time, req *jvspb.C
 	}
 
 	id := uuid.New().String()
+	if req.Ttl.AsDuration() > time.Duration(24*time.Hour) {
+		return nil, fmt.Errorf("token ttl should be less than 24 hours")
+	}
 	exp := now.Add(req.Ttl.AsDuration())
+
 	justs := req.Justifications
 
 	// Use audiences in the request if provided.
--- a/pkg/justification/processor.go
+++ b/pkg/justification/processor.go
@@ -164,7 +164,7 @@ func (p *Processor) createToken(ctx context.Context, now time.Time, req *jvspb.C
 	}
 
 	id := uuid.New().String()
-	if req.Ttl.AsDuration() > time.Duration(24*time.Hour) {
+	if req.Ttl.AsDuration() > 24*time.Hour {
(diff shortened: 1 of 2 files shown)


pkg/justification/processor_test.go
```diff
@@ -109,6 +109,13 @@ func TestCreateToken(t *testing.T) {
 			},
 			wantErr: "failed to validate request",
 		},
+		{
+			name: "ttl_too_long",
+			request: &jvspb.CreateJustificationRequest{
+				Ttl: durationpb.New(25 * time.Hour),
+			},
+			wantErr: "failed to validate request",
```
yolocs, line 117: Should we use the new error message?

pkg/justification/processor.go
```diff
@@ -164,7 +164,11 @@ func (p *Processor) createToken(ctx context.Context, now time.Time, req *jvspb.C
 	}
 
 	id := uuid.New().String()
+	if req.Ttl.AsDuration() > 24*time.Hour {
+		return nil, fmt.Errorf("token ttl should be less than 24 hours")
+	}
```
yolocs, line 169: Relocate to `runValidations`?

yolocs (changes requested):

pkg/justification/processor_test.go
capri-xiyue: +1 can we add more specific wantErr substring in UT?

pkg/justification/processor.go
```diff
@@ -78,8 +78,9 @@ func (p *Processor) CreateToken(ctx context.Context, req *jvspb.CreateJustificat
 	logger := logging.FromContext(ctx)
 
 	if err := p.runValidations(req); err != nil {
+		// panic(err.Error())
 		logger.Errorw("failed to validate request", "error", err)
-		return nil, status.Error(codes.InvalidArgument, "failed to validate request")
+		return nil, status.Error(codes.InvalidArgument, "failed to validate request, "+err.Error())
```
capri-xiyue, line 83: use`fmt.Sprintf()`

sethvargo:
Sorry hang on. I have a PR in progress for this, but was trying to land a few other things first. I think we want to do something a bit more sophisticated which I have coded up locally, but was running into issues testing (which Chen fixed in another PR).

sethvargo:
https://github.com/abcxyz/jvs/pull/176

Issue #170, opened 2023-01-24
Token reason should have a size limit (or with truncate)


yolocs:
### TL;DR

The size limit of a HTTP header is 8KB which marks our upper limit for the token size. So for the reason, we should limit to something much lower, like 2KB.

PR #177, opened 2023-01-25
Limit size on justifications

sethvargo:
This isn't a perfect solution because there's some error during JSON marshalling, but I don't think we should take the overhead of marshalling the JSON.

Fixes GH-170


sethvargo:
Ping @yolocs

yolocs (approved):

PR #183, opened 2023-02-21
Misc feedback on the UI flow

shan4993:
Addressing the following concerns brought up in a previous PR

1. use `querySelector` in place of `addEventListener` in the main script for the popup
2. handle NPE cases for the same script mentioned in 1
3. remove awkward if/else in the success page script, call window.close() to exit early
4. nit: address the spacing in the test classes
5. prefer `any` over `interface{}` in the `RenderHTMLStatus` method
6. change the `Server` struct to have a `Controller` then give the `Controller` struct the `allowlist` to avoid passing it as a param

### DIFF
--- a/assets/server/static/js/popup/main.js
+++ b/assets/server/static/js/popup/main.js
@@ -12,35 +12,40 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-const originElement = document.querySelector("#origin");
-const windowElement = document.querySelector("#windowname");
+window.addEventListener("DOMContentLoaded", async () => {
+  const originElement = document.querySelector("#origin");
+  const windowElement = document.querySelector("#windowname");
 
-if (!originElement) {
-  alert("The origin input was not detected.");
-  window.close();
-}
-
-if (!windowElement) {
-  alert("The windowname input was not detected.");
-  window.close();
-}
-
-// leverage the URL parameter provided from the request and set it to the target origin
-const encodedUriComponent = new URLSearchParams(window.location.search).get("origin");
-if (!encodedUriComponent && !originElement) {
-  alert("An origin URL parameter must be provided.");
-  window.close();
-}
-
-if (encodedUriComponent) {
-  const targetOrigin = decodeURIComponent(encodedUriComponent);
-  if (!targetOrigin) {
-    alert("Decoded URL parameter is invalid.");
+  if (!originElement) {
+    alert("The origin input was not detected.");
     window.close();
+    return;
   }
 
-  // set values for the following hidden input elements, will be persisted to the next page
-  originElement.value = targetOrigin;
-  windowElement.value = window.name;
-}
+  if (!windowElement) {
+    alert("The windowname input was not detected.");
+    window.close();
+    return;
+  }
 
+  // leverage the URL parameter provided from the request and set it to the target origin
+  const encodedUriComponent = new URLSearchParams(window.location.search).get("origin");
+  if (!encodedUriComponent && !originElement) {
+    alert("An origin URL parameter must be provided.");
+    window.close();
+    return;
+  }
+
+  if (encodedUriComponent) {
+    const targetOrigin = decodeURIComponent(encodedUriComponent);
+    if (!targetOrigin) {
+      alert("Decoded URL parameter is invalid.");
+      window.close();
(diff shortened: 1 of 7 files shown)


assets/server/static/js/popup/main.js
```diff
@@ -12,28 +12,35 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-window.addEventListener("DOMContentLoaded", async () => {
-  const originElement = document.getElementById("origin");
-  const windowElement = document.getElementById("windowname");
+const originElement = document.querySelector("#origin");
```
verbanicm, line 15: I dont think the goal was to remove addEventListener, rather to just switch from getElementById to querySelector?

sethvargo: Correct

assets/server/static/js/popup/main.js
```diff
@@ -12,28 +12,35 @@
...
+const windowElement = document.querySelector("#windowname");
 
-  // leverage the URL parameter provided from the request and set it to the target origin
-  const encodedUriComponent = new URLSearchParams(window.location.search).get("origin");
-  if (!encodedUriComponent && !originElement) {
-    alert("An origin URL parameter must be provided.")
+if (!originElement) {
+  alert("The origin input was not detected.");
+  window.close();
```
sethvargo, line 20: I think we need a `return` after all of these to prevent execution from continuing.

pkg/controller/controller.go
```diff
@@ -81,19 +82,20 @@ var (
...
 
-func (c *Controller) HandlePopup(allowlist []string) http.Handler {
+func (c *Controller) HandlePopup() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch r.Method {
 		case http.MethodGet:
 			c.handlePopupGet(w, r)
 		case http.MethodPost:
-			c.handlePopupPost(w, r, allowlist)
+			c.handlePopupPost(w, r, c.allowlist)
```
sethvargo, line 98: The allowlist is already dangled off the controller, so you don't need it as a parameter here. It's accessible within `handlePopupPost` off of `c`.

assets/server/static/js/popup/main.js
shan4993: ack

pkg/controller/controller.go
shan4993: ah of course

assets/server/static/js/popup/main.js
sethvargo: What error?

shan4993: by removing the method scope i was triggering the error, bringing the returns back.

verbanicm: I didn't follow this.

verbanicm: You caused the error when you removed the addEventListener but if you add that back the returns will work?

shan4993: yes, i think its because the the second argument to addEventListener is a method. So you can return inside that method. Without the addEventListener, return triggers an error: 

```
Uncaught SyntaxError: Illegal return statement
```

verbanicm: Right, adding that back with the return should be good.

assets/server/static/js/success/main.js
```diff
@@ -12,18 +12,30 @@
...
-const token = scriptTag.getAttribute("data-token");
-if (!targetOrigin) {
-  alert("You must pass a target origin from your application to successfully retrieve a token.")
-  window.close()
-} else if (!windowName) {
-  alert("You must pass a window name from your application to successfully retrieve a token.")
-} else if (!token) {
-  alert("Something went wrong, unable to retrieve a token.")
-} else {
+document.addEventListener("DOMContentLoaded", async () => {
```
verbanicm, line 15: Should we use window instead of document here also?

shan4993: Yes for consistency.

verbanicm:
LGTM, but I cant approve as Im not a writer.

sethvargo (approved):

bradegler (approved):

PR #185, opened 2023-02-23
Separate subject from requestor

sethvargo:
The current implementation is tightly coupled to incoming authentication being set on the gRPC request, which introduces two challenges:

1. There are situations where the entity minting the token is not the entity using the justification (e.g. a proxy, pair debugging). Per the design document, services need to verify that the JWT subject matches the caller subject; it's not the JVS' job to verify a caller's identity matches the identity of the justification. If someone mints a token for an incorrect email address, downstream clients should reject the request when the calling identity doesn't match the justification identity (subject).

2. There are many ways one might authenticate to a JVS (API, IAP, proxy, CLI, UI), and all of these have unique ways of interacting with the service.

This PR decouples the "metadata extraction" from "subject setting" and enables callers to mint JVS tokens with any provided subject.

- [Original JVS PRD](https://docs.google.com/document/d/1EpNyjDhijNa80VXG3Mhnr3X88kS_AodvH-SAtYQYy6g/edit?resourcekey=0-DGB7h6JEQUNMYVCFEsCP6g#heading=h.qav149qyfug)
- [Original JVS Design](https://docs.google.com/document/d/1DCRs0zwbg_8boxHorfmTgdVwU2cGjICTaYTuncvBhOA/edit#heading=h.5irk4csrpu0y)

### TODOs

- [x] Make sure I'm not crazy
- [x] Environment variable for the CLI subject (to save typing)
- [x] ~Java bits~


protos/v0/jvs_request.proto
```diff
@@ -28,6 +28,10 @@ message CreateJustificationRequest {
 
   // Optional audiences for the justifications.
   repeated string audiences = 3;
+
+  // Subject will tbe the corresponding subject in the JWT. Consumers of these
```
shan4993, line 32: nit: will be*

protos/v0/jvs_request.proto
```diff
@@ -28,6 +28,10 @@ message CreateJustificationRequest {
...
+  // JWTs must verified that the caller matches the embedded principal.
```
shan4993, line 33: nit: must verify*

shan4993 (approved):

test/integ/main_test.go
```diff
@@ -115,13 +108,7 @@ func TestJVS(t *testing.T) {
...
-	validJWT := testutil.SignToken(t, tok, authKey, keyID)
-	ctx = metadata.NewIncomingContext(ctx, metadata.New(map[string]string{
-		"authorization": "Bearer " + validJWT,
-	}))
-
-	p := justification.NewProcessor(kmsClient, cfg, authHandler)
+	p := justification.NewProcessor(kmsClient, cfg)
 	jvsAgent := justification.NewJVSAgent(p)
 
 	tests := []struct {
```
sqin2019, line 114: Missing subject in CreateJustificationRequest seems cause the ci failure.

sqin2019:
Java bits as in jvs client-lib, or somewhere else?

sethvargo:
> Java bits as in jvs client-lib, or somewhere else?

With the new approach, I don't think it's necessary anymore.

pkg/cli/token.go
```diff
@@ -83,7 +94,8 @@ For example:
 	flags := cmd.Flags()
 	flags.StringVarP(&opts.explanation, "explanation", "e", "",
 		"The explanation for the action")
-	cmd.MarkFlagRequired("explanation") //nolint // not expect err
+	flags.StringVarP(&opts.subject, "subject", "s", os.Getenv("JVSCTL_SUBJECT"),
+		"The principal that be making calls with the justification token")
```
shan4993, line 98: nit: "The principal that will make calls with the justification token" ?

shan4993 (approved):

pkg/cli/token.go
```diff
@@ -104,6 +116,10 @@ func runTokenCmd(cmd *cobra.Command, opts *tokenCmdOptions, args []string) error
 	ctx := context.Background()
 	out := cmd.OutOrStdout()
 
+	if opts.explanation == "" {
```
verbanicm, line 119: Should we validate the subject is not empty also? Just thinking of the case where someone didnt set the env var and didnt use the flag. Would be frustrating later when the token doesnt work.

sqin2019 (approved):

PR #186, opened 2023-02-24
Decouple authentication from jvsctl

sethvargo:
This marks authentication as optional and removes the automatic lookup of the local ID token. Users can still pass authentication using `-auth-token=$(gcloud auth print-identity-token)`, but this removes much of the auth lookup magic.


capri-xiyue:
@sqin2019  Is this a break change to our existing users?

sethvargo:
I don't think so, because I don't think we've released a version of the JVS with `jvsctl` as a binary yet.

sqin2019 (approved):

Issue #187, opened 2023-02-27
SDK validation func should accept an expected subject to match the sub claim


yolocs:
### TL;DR

If the subject doesn't match, the token validation should fail.
Validating breakglass token should preserve its current logic.

yolocs:
cc: @sethvargo

PR #191, opened 2023-02-28
Update SDK to validate JWT subject

sethvargo:
Fixes GH-187

### DIFF
--- a/client-lib/java/src/main/java/com/abcxyz/jvs/JvsClient.java
+++ b/client-lib/java/src/main/java/com/abcxyz/jvs/JvsClient.java
@@ -99,7 +99,7 @@ public class JvsClient {
       Algorithm algorithm = Algorithm.ECDSA256((ECPublicKey) jwk.getPublicKey(), null);
       algorithm.verify(token);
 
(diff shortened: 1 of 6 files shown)


sethvargo:
The integration tests seem to be failing for a different reason.

client-lib/java/src/main/java/com/abcxyz/jvs/JvsClient.java
```diff
@@ -98,6 +98,15 @@ public DecodedJWT validateJWT(String tokenStr) throws JwkException {
       }
       Algorithm algorithm = Algorithm.ECDSA256((ECPublicKey) jwk.getPublicKey(), null);
       algorithm.verify(token);
+
+      if (!expectedSubject.equals(token.getSubject()) && !expectedSubject.equals("")) {
```
yolocs, line 102:
```suggestion
      if (!expectedSubject.equals(token.getSubject()) && !expectedSubject.isBlank()) {
```

yolocs (approved):

client-lib/java/src/main/java/com/abcxyz/jvs/JvsClient.java
sethvargo: Done

PR #192, opened 2023-03-01
create token on form submission

shan4993:
Leverage a processor in the ui's controller to mint tokens. Also using an IAP header to reference the requestor.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -159,9 +159,7 @@ func (c *Controller) handlePopupPost(w http.ResponseWriter, r *http.Request) {
 				Value:    formDetails.Reason,
 			},
 		},
-		Ttl: &durationpb.Duration{
-			Seconds: int64(dur.Seconds()),
-		},
+		Ttl: durationpb.New(dur),
 	}
 
 	token, err := c.p.CreateToken(context.Background(), formDetails.UserEmail, req)
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -15,20 +15,26 @@
 package controller
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"net/http"
 	"net/url"
+	"strconv"
 	"strings"
 
+	jvspb "github.com/abcxyz/jvs/apis/v0"
 	"github.com/abcxyz/jvs/internal/project"
+	"github.com/abcxyz/jvs/pkg/justification"
 	"github.com/abcxyz/jvs/pkg/render"
 	"golang.org/x/exp/slices"
+	"google.golang.org/protobuf/types/known/durationpb"
 )
 
 // Controller manages use of the renderer in the http handler.
 type Controller struct {
 	h         *render.Renderer
+	p         *justification.Processor
 	allowlist []string
 }
 
@@ -54,6 +60,7 @@ type FormDetails struct {
 	Origin      string
 	PageTitle   string
 	Description string
+	UserEmail   string
 	Content     Content
 	Category    string
 	Reason      string
@@ -82,9 +89,10 @@ var (
 	ttls       = []string{"15", "30", "60", "120", "240"}
 )
 
-func New(h *render.Renderer, allowlist []string) *Controller {
+func New(h *render.Renderer, p *justification.Processor, allowlist []string) *Controller {
 	return &Controller{
 		h:         h,
+		p:         p,
 		allowlist: allowlist,
 	}
 }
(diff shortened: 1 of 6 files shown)


pkg/controller/controller.go
```diff
@@ -143,14 +145,35 @@ func (c *Controller) handlePopupPost(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// 3. [TODO] Request a token
-	token := "token_from_server"
+	// 3. Request a token
+	ttl, err := strconv.Atoi(formDetails.TTL)
```
sethvargo, line 149: We should accept ttls as strings (e.g. "12s") and use `time.ParseDuration`.

pkg/controller/controller.go
```diff
@@ -143,14 +145,35 @@ func (c *Controller) handlePopupPost(w http.ResponseWriter, r *http.Request) {
...
+			},
+		},
+		Ttl: &durationpb.Duration{
+			Seconds: int64(ttl) * 60,
+		},
+	}
+
+	token, err := c.p.CreateToken(context.Background(), formDetails.UserEmail, req)
+	if err != nil {
+		c.renderBadRequest(w, err.Error())
```
sethvargo, line 169: You need a `return` after this

pkg/controller/controller.go
```diff
@@ -143,14 +145,35 @@ func (c *Controller) handlePopupPost(w http.ResponseWriter, r *http.Request) {
...
+	}
 
 	// 4. Redirect to a confirmation page with context, ultimately needed to postMessage back to the client
 	successDetails := &SuccessDetails{
 		PageTitle:   "JVS - Successful token retrieval",
 		Description: "Successful token page",
-		Token:       token,
+		Token:       string(token),
```
sethvargo, line 176: Is `token` a `[]byte`?

pkg/controller/controller.go
```diff
@@ -289,3 +313,13 @@ func getFormDetails(r *http.Request) *FormDetails {
 		},
 	}
 }
+
+func (c *Controller) renderBadRequest(w http.ResponseWriter, m string) {
```
sethvargo, line 317: document

pkg/controller/controller.go
```diff
@@ -289,3 +313,13 @@ func getFormDetails(r *http.Request) *FormDetails {
...
+	t := http.StatusText(http.StatusBadRequest)
+	forbiddenDetails := &ErrorDetails{
+		PageTitle:   t,
+		Description: t,
+		Message:     m,
+	}
+	c.h.RenderHTMLStatus(w, http.StatusBadRequest, "400.html.tmpl", forbiddenDetails)
```
sethvargo, line 324: You can inline this:

```suggestion
	c.h.RenderHTMLStatus(w, http.StatusBadRequest, "400.html.tmpl", &ErrorDetails{
		PageTitle:   t,
		Description: t,
		Message:     m,
	})
```

pkg/controller/controller.go
shan4993: yes it is

pkg/controller/controller.go
shan4993: cool didnt know about this library, this simplifies things.

pkg/controller/controller.go
```diff
@@ -143,14 +145,36 @@ func (c *Controller) handlePopupPost(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	req := &jvspb.CreateJustificationRequest{
+		Justifications: []*jvspb.Justification{
+			{
+				Category: formDetails.Category,
+				Value:    formDetails.Reason,
+			},
+		},
+		Ttl: &durationpb.Duration{
```
sethvargo, line 162: `durationpb.New(dur)`

pkg/controller/controller.go
```diff
@@ -79,12 +86,13 @@ type ErrorDetails struct {
 
 var (
 	categories = []string{"explanation", "breakglass"}
-	ttls       = []string{"15", "30", "60", "120", "240"}
+	ttls       = []string{"15m", "30m", "1h", "2h", "4h"}
```
sethvargo, line 89: This is more general advice than a specific change request. Slices (and maps) are mutable in Go, so this kind of design lends itself to some difficult-to-debug edge cases: https://go.dev/play/p/A7JtU5oznsT. See "A possible gotcha" in https://go.dev/blog/slices-intro.

For this reason, especially for smaller structures like there, I recommend having a function that returns a copy:

```go
func ttls() []string {
  return []string{"15m", ...}
}
```

Note, do _not_ do this. Since slices are passed by reference, this is the same as just having the variable:

```go
# NO, BAD
var _ttls = []string{"15m", ...}

func ttls() []string {
  return ttls
}
```

pkg/ui/server.go
```diff
@@ -32,15 +35,27 @@ type Server struct {
...
+		return nil, fmt.Errorf("failed to setup kms client: %w", err)
+	}
+
+	var cfg config.JustificationConfig
+	if err := cfgloader.Load(ctx, &cfg); err != nil {
+		return nil, fmt.Errorf("failed to load config: %w", err)
+	}
+
+	p := justification.NewProcessor(kmsClient, &cfg)
+
```
sethvargo, line 56: I think it would make more sense to do this higher up the stack (in main.go) and then update the signature for `NewServer` to accept a processor.

shan4993: ack on it

pkg/controller/controller.go
shan4993: oh thanks for sharing, ill update for this

PR #193, opened 2023-03-03
cleanup: terraform modules and prepare them for better integration tests

yolocs:
Changes:

- Merged all service modules into one `modules/jvs-services`. And used https://github.com/abcxyz/terraform-modules/tree/main/modules/cloud_run to create services.
- Relocated KMS keyring and service accounts to a module `modules/common`. These are resources are more "static". Unlike services, these resources usually only need to be created once.
- Revamped `ci` module to use https://github.com/abcxyz/terraform-modules/tree/main/modules/github_ci_infra and call `modules/common` to create "static" resources.
  - In the near future, each CI run will deploy new `jvs-services` to test the new changes with dedicated new services. This allows us to run multiple CIs in parallel.
  - This change will be a breaking change because https://github.com/abcxyz/terraform-modules/tree/main/modules/github_ci_infra will create a different CI service account. After applying the terraform, I need to update the JVS workflow to use the new CI service account.

TODO:

- We will want a terraform module that can bring up the full end-to-end JVS setup: issue


terraform/modules/jvs-services/jvs-ui.tf
```diff
@@ -0,0 +1,30 @@
...
+# limitations under the License.
+
+module "ui_cloud_run" {
+  source = "git::https://github.com/abcxyz/terraform-modules.git//modules/cloud_run?ref=5445543e21491176528fb5cd7adcb505d9dec5dd"
+
+  project_id            = var.project_id
+  region                = var.region
+  name                  = var.jvs_ui_service_name
+  image                 = var.jvs_ui_service_image
+  ingress               = var.service_ingress
```
yolocs, line 22: @shan4993 By default, I'm using `all` ingress assuming IAP is not in the picture. I suppose the future integration test can carry out without IAP.

In the e2e setup, I plan to introduce gclb and IAP https://github.com/abcxyz/jvs/issues/194.

terraform/ci/main.tf
```diff
@@ -14,75 +14,37 @@
...
 locals {
-  github_slug = "abcxyz/jvs"
-}
-
-resource "google_project_service" "serviceusage" {
-  project            = var.project_id
-  service            = "serviceusage.googleapis.com"
-  disable_on_destroy = false
+  github_owner_id = 93787867  # abcxyz
+  github_repo_id  = 479173136 # jvs
```
sethvargo, line 18:
```suggestion
  github_repo_id  = 479173136 # abcxyz/jvs
```

terraform/modules/jvs-services/public-key.tf
```diff
@@ -0,0 +1,29 @@
...
+  project_id            = var.project_id
+  region                = var.region
+  name                  = var.jvs_public_key_service_name
+  image                 = var.jvs_public_key_service_image
+  ingress               = var.service_ingress
+  service_account_email = var.jvs_public_key_service_account
+  service_iam           = var.jvs_public_key_service_iam
+  envvars = {
+    "KEY_NAMES" : google_kms_crypto_key.signing_key.id,
+    "CACHE_TIMEOUT" : var.key_cache_timeout,
```
sethvargo, line 27: Can be a followup, but let's prefer "KMS_*" for KMS-related configs and "JVS_*" for JVS-related configs.

terraform/modules/jvs-services/variables.tf
```diff
@@ -0,0 +1,213 @@
...
+  type        = number
+  default     = 5
+  description = "Cadence (expressed in minutes) to run the certificate rotator on. If set to 0, key rotation won't be scheduled."
+}
+
+variable "key_cache_timeout" {
+  type        = string
+  default     = "10m"
+  description = "Duration before cache entries are invalided."
+}
```
sethvargo, line 201: Can we make these all `kms_*`? This can be a follow-up.

terraform/modules/jvs-services/variables.tf
```diff
@@ -0,0 +1,213 @@
...
+  description = "length of time that it takes for a change in the key in KMS to be reflected in the client."
+  type        = string
+  default     = "2m"
+}
+
+variable "key_rotation_cadence" {
+  type        = number
+  default     = 5
+  description = "Cadence (expressed in minutes) to run the certificate rotator on. If set to 0, key rotation won't be scheduled."
+}
```
sethvargo, line 195: Per the evolving Terraform style guide, this should be "key_rotation_minutes": [go/bets-terraform-guide](https://goto.google.com/bets-terraform-guide)

terraform/modules/jvs-services/variables.tf
```diff
@@ -0,0 +1,213 @@
...
+    admins     = list(string)
+    developers = list(string)
+    invokers   = list(string)
+  })
+  default = {
+    admins     = []
+    developers = []
+    invokers   = []
+  }
+}
```
sethvargo, line 61: I see this pattern repeated multiple times (service name, service account, service image, service iam), and it feels a little strange. Is there a reason these need to be variables?

terraform/modules/jvs-services/variables.tf
```diff
@@ -0,0 +1,213 @@
...
+  type        = string
+  default     = "false"
+  description = "Whether UI is in dev mode."
+}
+
+variable "ui_origin_allowlist" {
+  type        = string
+  default     = "*"
+  description = "UI origin allowlist."
+}
```
sethvargo, line 213: Aren't these just environment variables to the service? Can we have environment variables as a variable and set some sane defaults? This is what we did in EN; I don't want to have to mirror every app-level configurable as a Terraform variable.

sethvargo (changes requested):

terraform/ci/main.tf
```diff
@@ -14,75 +14,37 @@
...
 }
 
 resource "google_project_service" "services" {
   project = var.project_id
   for_each = toset([
```
bradegler, line 23: Please put the for_each above the project line

terraform/modules/common/main.tf
```diff
@@ -0,0 +1,98 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+resource "google_project_service" "services" {
+  project = var.project_id
+  for_each = toset([
```
bradegler, line 17: for_each about project here too please.

terraform/modules/common/main.tf
```diff
@@ -0,0 +1,98 @@
...
+}
+
+resource "google_kms_key_ring_iam_member" "rotator_acc_roles" {
+  for_each = toset([
+    "roles/cloudkms.admin",
+  ])
+
+  key_ring_id = google_kms_key_ring.keyring.id
+  role        = each.key
+  member      = "serviceAccount:${google_service_account.rotator_acc.email}"
```
bradegler, line 80: You could simplify this

```suggestion
  member      = google_service_account.rotator_acc.member
```

terraform/modules/common/main.tf
```diff
@@ -0,0 +1,98 @@
...
+
+resource "google_kms_key_ring_iam_member" "public_key_acc_roles" {
+  for_each = toset([
+    "roles/cloudkms.publicKeyViewer",
+    "roles/cloudkms.viewer",
+  ])
+
+  key_ring_id = google_kms_key_ring.keyring.id
+  role        = each.key
+  member      = "serviceAccount:${google_service_account.public_key_acc.email}"
```
bradegler, line 97:
```suggestion
  member      = google_service_account.public_key_acc.member
```

bradegler (changes requested):

terraform/modules/jvs-services/public-key.tf
yolocs: You mean for environment variables?

terraform/modules/jvs-services/variables.tf
yolocs: At least half of them are required:

- Service accounts are created in a separate module so we need to pass them over
- Images are built separately so we need to pass them over

The service IAM and name are optional. For simplicity, I can get rid of them for now.

terraform/modules/jvs-services/variables.tf
yolocs: Done

terraform/modules/jvs-services/variables.tf
yolocs: Removed IAM and service name

terraform/modules/jvs-services/variables.tf
yolocs: Changed to directly having env vars as input.

bradegler (approved):

terraform/modules/jvs-services/jvs-ui.tf
shan4993: Sounds good. The future integ test would need that email header from IAP but since we arent relying on the JWT header i imagine we can set a random value via integ test.

yolocs: Yes, that's what I was thinking.

terraform/ci/main.tf
```diff
@@ -14,75 +14,38 @@
...
   role     = each.key
-  member   = "serviceAccount:${google_service_account.gh-access-acc.email}"
+  member   = module.github_ci_infra.service_account_member
 }
 
-module "workload-identity-federation" {
-  source      = "github.com/abcxyz/pkg//terraform/modules/workload-identity-federation"
-  project_id  = var.project_id
-  github_slug = local.github_slug
+module "github_ci_infra" {
```
sqin2019, line 39: Should this belong to infra pkg?

shan4993 (approved):

terraform/modules/jvs-services/public-key.tf
sethvargo: Yea

sethvargo (approved):
I still have an open comment that we can discuss later. I don't like the repetition of 5 variables per service.

terraform/ci/main.tf
sqin2019: discussed offline, resolving.

terraform/modules/jvs-services/public-key.tf
yolocs: https://github.com/abcxyz/jvs/issues/195

terraform/modules/jvs-services/variables.tf
yolocs: Ack. Tracking in https://github.com/abcxyz/jvs/issues/196

terraform/ci/main.tf
yolocs: Discussed offline. If this is a new project, I'd make it in the infra repo.

PR #199, opened 2023-03-03
Switch to pkg/renderer

sethvargo:
github.com/abcxyz/pkg#73 ported the renderer from this project into pkg. This moves jvs to use the pkg renderer.


shan4993 (approved):
Thanks Seth!

bradegler (approved):

PR #202, opened 2023-03-06
fix: parse out the email from IAP

shan4993:


pkg/controller/controller.go
```diff
@@ -84,6 +84,8 @@ type ErrorDetails struct {
 	Message     string
 }
 
+var iapHeaderName = "x-goog-authenticated-user-email"
```
yolocs, line 87: Nit: const?

pkg/controller/controller.go
```diff
@@ -298,6 +313,19 @@ func (c *Controller) renderBadRequest(w http.ResponseWriter, m string) {
...
 }
 
+func getEmail(iapEmailValue string) (string, error) {
+	if iapEmailValue == "" {
+		return "", fmt.Errorf("email header is not valid")
+	}
+
+	split := strings.Split(iapEmailValue, ":")
+	if len(split) != 2 {
+		return "", fmt.Errorf("email value has unexpected format")
```
yolocs, line 323: Care to add want vs. got?

pkg/controller/controller_test.go
```diff
@@ -376,3 +373,46 @@ func TestIsValidOneOf(t *testing.T) {
...
+		tc := tc
+
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			gotRes, err := getEmail(tc.email)
+			if diff := testutil.DiffErrString(err, tc.wantErr); diff != "" {
+				t.Errorf("Unexpected err: %s", diff)
+			}
+			if diff := cmp.Diff(tc.wantRes, gotRes); diff != "" {
```
yolocs, line 413: Use `!=` for primitive type comparison

yolocs (changes requested):

internal/envtest/envtest.go
```diff
@@ -48,7 +48,7 @@ type ServerConfigResponse struct {
 // BuildFormRequest builds an http request and http response recorder for the
 // given form values (expressed as url.Values). It sets the proper headers and
 // response types to post as a form and expect HTML in return.
-func BuildFormRequest(ctx context.Context, tb testing.TB, meth, pth string, v *url.Values) (*httptest.ResponseRecorder, *http.Request) {
+func BuildFormRequest(ctx context.Context, tb testing.TB, meth, pth string, v *url.Values, customHeaders *http.Header) (*httptest.ResponseRecorder, *http.Request) {
```
sethvargo, line 51: I'm not sure I like this change. The builder returns the `http.Request` for this kind of customization on purpose. I would push this into the test itself:

```go
w, r := BuildFormRequest(ctx, t, "GET", "/", nil)
r.Header.Set("foo", "bar")
```

pkg/controller/controller.go
```diff
@@ -298,6 +313,19 @@ func (c *Controller) renderBadRequest(w http.ResponseWriter, m string) {
 	})
 }
 
+func getEmail(iapEmailValue string) (string, error) {
```
sethvargo, line 316: Nit: I would expect this to accept an `r *http.Request` instead. Shouldn't change much about the unit testing, but it provides a stronger encapsulation.

pkg/controller/controller.go
sethvargo, line 87: Should we be validating these headers? https://cloud.google.com/iap/docs/signed-headers-howto

sethvargo (changes requested):

pkg/controller/controller.go
shan4993: Yeah I was thinking about this, I decided not to for the following reasons:

1. Given your recent decoupling of the email for subject and letting the downstream service validate reduces the emphasis on the validity of the email 
2. AFIK local testing wouldnt be possible given we would need a valid JWT from IAP? [They have this](https://cloud.google.com/iap/docs/query-parameters-and-headers-howto#testing_jwt_verification) for testing but its through query params instead of headers.

pkg/controller/controller.go
yolocs: Yes, expected format. That's what I mean.

shan4993: i believe i got it to work 

```
	got, want := strings.Split(iapEmailValue, ":"), []string{iapHeaderName + "<email>"}
	if len(got) != 2 {
		return "", fmt.Errorf("email %q does not match expected email pattern %q", got, want)
	}
```

pkg/controller/controller.go
yolocs: I'm fine not doing it in this PR (I thought we were going to add validation earlier) but we should prioritize adding it.

pkg/controller/controller_test.go
shan4993: Im probably overthinking this, could you provide an example? Im thinking if I do != then I wont have a useful error message to print.

pkg/controller/controller.go
sethvargo: I agree this complicates local testing. I also agree with Chen is should be in a different PR. Let's make an issue to design out validation for gRPC and IAP headers, documenting the need for a good local dev story (e.g. maybe we skip validation when `DEV_MODE=true`).

shan4993: Ack ill create an issue.

pkg/controller/controller_test.go
```diff
@@ -86,7 +87,13 @@ func TestHandlePopup(t *testing.T) {
 			c := New(harness.Renderer, harness.Processor, tc.allowlist)
 
 			w, r := envtest.BuildFormRequest(ctx, t, tc.method, tc.path,
-				tc.queryParam, tc.headers)
+				tc.queryParam)
+
+			for key, values := range *tc.headers {
```
sethvargo, line 92: `http.Header` is a `map[string][]string`, which is already a pointer. The test cases should use `http.Header`, not `*http.Header`.

shan4993: ack i suppose the same is true for `url.Values`

pkg/controller/controller.go
yolocs: gRPC is tricky if it's on cloud run. Cloud run will redact the JWT signature so we can no longer validate it.

shan4993: https://github.com/abcxyz/jvs/issues/203

sethvargo: @yolocs I think they fixed that actually!

yolocs: Interesting! Do you have a reference? Haven't tried it for real but the doc still says so https://cloud.google.com/run/docs/troubleshooting#signature-removed

docs/jvs-ui.md
```diff
@@ -2,7 +2,7 @@
 
 **JVS is not an official Google product.**
 
-[JVS UI](../cmd/ui) facilitates the justification verification flow using a UI
+[JVS UI](../cmd/ui) facilitates the justification verification flow using a UI. It is meant to be an alternative to using `jvsctl` which requires integrating a calling application that is non-customer facing.
```
yolocs, line 5: It's probably better to have this in a separate PR

pkg/controller/controller_test.go
yolocs: You just write it yourself

```go
if tc.wantRes != gotRest {
  t.Errorf("email got=%s want=%s", gotRes, tc.wantRes)
}
```

docs/jvs-ui.md
shan4993: yeah didnt mean to bring it in yet, redacting until we get those screenshots.

pkg/controller/controller.go
sethvargo: We might be talking about two different things. I'll ping you on chat.

pkg/controller/controller_test.go
sethvargo: Even better, shadow the variable to make the error message and format consistent:

```go
if got, want := gotRest, tc.wantRes; got != want {
  t.Errorf("expected %q to be %q", got, want)
}
```

pkg/controller/controller.go
```diff
@@ -298,6 +313,21 @@ func (c *Controller) renderBadRequest(w http.ResponseWriter, m string) {
 	})
 }
 
+func getEmail(r *http.Request) (string, error) {
+	iapEmailValue := r.Header.Get(iapHeaderName)
+
+	if iapEmailValue == "" {
+		return "", fmt.Errorf("email header is not valid")
```
sethvargo, line 320:
```suggestion
		return "", fmt.Errorf("email header is not present")
```

pkg/controller/controller.go
```diff
@@ -298,6 +313,21 @@ func (c *Controller) renderBadRequest(w http.ResponseWriter, m string) {
...
+	}
+
+	split := strings.Split(iapEmailValue, ":")
+	if len(split) != 2 {
+		return "", fmt.Errorf("email value has unexpected format, expected %s:<email>", iapHeaderName)
+	}
```
sethvargo, line 326: https://cloud.google.com/iap/docs/signed-headers-howto#retrieving_the_user_identity says:

> Unlike that header and the `sub` claim, this value doesn't have a namespace prefix.

I don't think there's a colon split in this value.

shan4993: This is referring to the JWT post validation. We're actually using the other email header. Part of the issue I created addresses using the JWT from IAP.

yolocs (approved):

capri-xiyue (approved):

sethvargo (approved):

PR #206, opened 2023-03-07
fix: release docker tag

yolocs:


.github/workflows/release.yml
```diff
@@ -122,22 +121,25 @@ jobs:
         service_account: '${{ env.WIF_SERVICE_ACCOUNT }}'
     - name: 'Deploy Cloud Run services'
       run: |-
+        # Image tags are without 'v' prefix.
+        DOCKER_TAG=$(echo '${{ github.ref_name }}' | cut -c 2-)
```
sethvargo, line 125: This assumes the string starts with a `v`. It'd be better to do:

```bash
DOCKER_TAG="${{ github.ref_name }}"
DOCKER_TAG=${DOCKER_TAG#v}"
```

capri-xiyue (approved):

PR #222, opened 2023-03-22
Drop cobra and viper

sethvargo:
This also switches to using environment variables instead of relying on a YAML file for configuration. Users can still set environment variables in their shellrc file, but we avoid tightly coupling `jvsctl` to a particular config format.

Depends on https://github.com/abcxyz/pkg/pull/100


docs/cli.md
```diff
@@ -12,7 +12,7 @@ go install github.com/abcxyz/jvs/cmd/jvsctl
 ```
 
 Or download from a
-[release](https://github.com/abcxyz/lumberjack/releases/tag/v0.0.4).
+[release](https://github.com/abcxyz/lumberjack/releases).
```
bradegler, line 15: This should be JVS, right?

```suggestion
[release](https://github.com/abcxyz/jvs/releases).
```

bradegler (changes requested):

bradegler (approved):

yolocs (approved):
Once we update de

PR #229, opened 2023-03-29
fix: update doc for jvscli

sailorlqh:
JVSCTL_SERVER_ADDRESS doesn't need the `https://` prefix. And the port by default is 443.
JVSCTL_JWKS_ENDPOINT doesn't the port number.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -28,13 +28,13 @@ the same justification server for minting tokens, add the following to your
 `.bashrc` or `.zshrc` file:
 
 ```shell
-export JVSCTL_SERVER_ADDRESS="https://jvs.corp.internal:8080"
+export JVSCTL_SERVER_ADDRESS="jvs.corp.internal:443"
 ```
 
 Similarly, you can set the endpoint for getting the JWKS for verification:
 
 ```shell
-export JVSCTL_JWKS_ENDPOINT="https://keys.corp.internal:8080/.well-known/jwks"
+export JVSCTL_JWKS_ENDPOINT="https://keys.corp.internal/.well-known/jwks"
 ```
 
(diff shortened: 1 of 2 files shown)


docs/cli.md
```diff
@@ -28,13 +28,13 @@ the same justification server for minting tokens, add the following to your
...
 ```shell
-export JVSCTL_SERVER_ADDRESS="https://jvs.corp.internal:8080"
+export JVSCTL_SERVER_ADDRESS="jvs.corp.internal:443"
 ```
 
 Similarly, you can set the endpoint for getting the JWKS for verification:
 
 ```shell
-export JVSCTL_JWKS_ENDPOINT="https://keys.corp.internal:8080/.well-known/jwks"
+export JVSCTL_JWKS_ENDPOINT="https://keys.corp.internal/.well-known/jwks"
```
sethvargo, line 37: These defaults intentionally match the default ports set in the config: https://github.com/abcxyz/jvs/blob/61f3158c299259a73471b9b0b5616f06638b2573/pkg/config/public_key_config.go#L32

docs/cli.md
```diff
@@ -28,13 +28,13 @@ the same justification server for minting tokens, add the following to your
 `.bashrc` or `.zshrc` file:
 
 ```shell
-export JVSCTL_SERVER_ADDRESS="https://jvs.corp.internal:8080"
+export JVSCTL_SERVER_ADDRESS="jvs.corp.internal:443"
```
sethvargo, line 31: 8080 is the correct default: https://github.com/abcxyz/jvs/blob/61f3158c299259a73471b9b0b5616f06638b2573/pkg/config/justification_config.go#L36

sethvargo (changes requested):

PR #239, opened 2023-04-07
feat: ci infra in workflow

sqin2019:
Was trying to verify past unverified commits and failed https://github.com/abcxyz/jvs/pull/233 ...so open a new one


yolocs (approved):
I assume it's the same content as the previous PR?

sqin2019:
> I assume it's the same content as the previous PR?

yes!

PR #240, opened 2023-04-07
Feat: add prober script and dockerfiles, update goreleaser.docker.yaml to build prober image

sailorlqh:


pkg/prober/prober.dockerfile
```diff
@@ -0,0 +1,15 @@
+
```
yolocs, line 1: Don't add non-go code to `pkg/`. I'm thinking to have `prober` folder at the repo root.

sailorlqh: Done. Moved to repo root.

.goreleaser.docker.yaml
```diff
@@ -280,6 +303,29 @@ dockers:
       - '--label=org.opencontainers.image.title=jvs-ui'
       - '--label=org.opencontainers.image.version={{ .Version }}'
 
+  -
+    ids:
```
capri-xiyue, line 307: Not that familiar with goreleaser. But wondering should we add this into the original jvs go releaser yaml?

.goreleaser.docker.yaml
```diff
@@ -115,6 +115,29 @@ builds:
...
+    main: ./cmd/jvsctl
+    binary: jvsctl
+    mod_timestamp: '{{ .CommitTimestamp }}'
+    flags:
+      - '-a'
+      - '-trimpath'
+    ldflags:
+      - '-s'
+      - '-w'
+      - '-X={{ .ModulePath }}/internal/version.Name=ui'
```
sqin2019, line 129: nit: name

prober/prober.sh
```diff
@@ -0,0 +1,11 @@
+set -eEuo pipefail
+
+ID_TOKEN=$(curl "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/identity?audience=$AUDIENCE" -H "Metadata-Flavor: Google")
+
+./jvsctl token --auth-token $ID_TOKEN -e "prober" > jvs_token
```
sqin2019, line 5: How do we pass server and jwks endpoint?

sailorlqh: It will be set in the configuration in the cloud run job as environment variables.

.goreleaser.docker.yaml
```diff
@@ -280,6 +303,29 @@ dockers:
...
       - '--label=org.opencontainers.image.version={{ .Version }}'
 
+  -
+    ids:
+    - jvsctl
+    use: 'buildx'
+    goos: 'linux'
+    goarch: 'amd64'
+    image_templates:
+      - '{{ .Env.REGISTRY }}/prober:{{ .Version }}-amd64'
```
sqin2019, line 313: FYI, you might need to resolve conflict since I merged https://github.com/abcxyz/jvs/pull/239, also please use DOCKER_TAG and .VERSION as a default, ```{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}```

sailorlqh: I resolved the conflict, but forgot this. Just made the change. Thanks for the note!

.goreleaser.docker.yaml
sailorlqh: Done.

prober/prober.dockerfile
```diff
@@ -0,0 +1,15 @@
+
```
sqin2019, line 1: nit: header.

prober/prober.sh
```diff
@@ -0,0 +1,11 @@
...
+
+cat jvs_token | ./jvsctl validate --token -
+
+./jvsctl token --auth-token $ID_TOKEN -e "prober" > jvs_breakglass_token
+
+cat jvs_breakglass_token | ./jvsctl validate --token -
```
sqin2019, line 11: Do we need to check the output and log the error?

prober/prober.sh
sqin2019, line 5: Can we add some error logs?

sailorlqh: During our previous discussion with Chen, our idea is if the script failed, there will be a corresponding error log automatically written by cloud run job which will contain the specific error message. Therefore, there is no need to right error logs here.

.goreleaser.docker.yaml
```diff
@@ -280,6 +303,52 @@ dockers:
...
+      - '--label=org.opencontainers.image.created={{ .CommitTimestamp }}'
+      - '--label=org.opencontainers.image.description=Prober service for jvs'
+      - '--label=org.opencontainers.image.licenses=Apache-2.0'
+      - '--label=org.opencontainers.image.name=prober'
+      - '--label=org.opencontainers.image.revision={{ .FullCommit }}'
+      - '--label=org.opencontainers.image.source={{ .GitURL }}'
+      - '--label=org.opencontainers.image.title=prober'
+      - '--label=org.opencontainers.image.version={{ .Version }}'
+
 docker_manifests:
```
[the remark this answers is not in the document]
sailorlqh: Thanks for the suggestions. I think currently we can just use this to fit the style with other targets. And later use a separate PR to update them all together. And focus this PR on add new features.

prober/prober.sh
sqin2019: Just wondering where will we check the validate output?

sailorlqh: Due to the screenshot contains internal info, this is explained via chat.

prober/prober.sh
```diff
@@ -0,0 +1,11 @@
+set -eEuo pipefail
+
+ID_TOKEN=$(curl "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/identity?audience=$AUDIENCE" -H "Metadata-Flavor: Google")
```
sethvargo, line 3:
```suggestion
ID_TOKEN=$(curl -sF "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/identity?audience=$AUDIENCE" -H "Metadata-Flavor: Google")
```

prober/prober.sh
sethvargo: https://github.com/abcxyz/pmap/tree/main/cmd/policy

prober/prober.dockerfile
```diff
@@ -0,0 +1,15 @@
...
+# Normally we would set this to run as "nobody".
+# But goreleaser builds the binary locally and sometimes it will mess up the permission
+# and cause "exec user process caused: permission denied".
+#
+# USER nobody
+
+# Run the web service on container startup.
+
+ENTRYPOINT ["bash", "prober.sh"]
```
yolocs, line 15: What will happen without `bash` here? Can't we use `prober.sh` as the entry point?

prober/prober.dockerfile
```diff
@@ -0,0 +1,15 @@
+# Base image that supports bash
+FROM cgr.dev/chainguard/bash:latest
+
+COPY jvsctl /jvsctl
+COPY ./prober/prober.sh prober.sh
```
yolocs, line 5: nit: to be consistent
```suggestion
COPY ./prober/prober.sh /prober.sh
```

prober/prober.sh
```diff
@@ -0,0 +1,11 @@
...
+
+./jvsctl token --auth-token $ID_TOKEN -e "prober" > jvs_token
+
+cat jvs_token | ./jvsctl validate --token -
+
+./jvsctl token --auth-token $ID_TOKEN -e "prober" > jvs_breakglass_token
```
capri-xiyue, line 9: Will the script continue to run the breakglass CUJ if the previous normal CUJ fail?

prober/prober.sh
yolocs: The CLI's output is the error log

prober/prober.sh
capri-xiyue: Can you add a summary here? I'm also wondering how you check the output?For example, we give a valid token but jvsctl returns status code 200 but said it's not valid, this also means there is something wrong with JVS service. Will we be able to catch it?

.goreleaser.docker.yaml
```diff
@@ -280,6 +303,52 @@ dockers:
...
+    image_templates:
+      - '{{ .Env.CONTAINER_REGISTRY }}/prober:{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}-amd64'
+    dockerfile: ./prober/prober.dockerfile
+    extra_files:
+    - ./prober/prober.sh
+    build_flag_templates:
+      - '--platform=linux/amd64'
+      - '--pull'
+      - '--label=org.opencontainers.image.created={{ .CommitTimestamp }}'
+      - '--label=org.opencontainers.image.description=Prober service for jvs'
```
yolocs, line 321: Not really a service
```suggestion
      - '--label=org.opencontainers.image.description=Prober for jvs'
```

.goreleaser.docker.yaml
```diff
@@ -280,6 +303,52 @@ dockers:
...
+    image_templates:
+      - '{{ .Env.CONTAINER_REGISTRY }}/prober:{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}-arm64'
+    dockerfile: ./prober/prober.dockerfile
+    extra_files:
+    - ./prober/prober.sh
+    build_flag_templates:
+      - '--platform=linux/amd64'
+      - '--pull'
+      - '--label=org.opencontainers.image.created={{ .CommitTimestamp }}'
+      - '--label=org.opencontainers.image.description=Prober service for jvs'
```
yolocs, line 344:
```suggestion
      - '--label=org.opencontainers.image.description=Prober for jvs'
```

.goreleaser.docker.yaml
```diff
@@ -280,6 +303,52 @@ dockers:
...
+      - '--label=org.opencontainers.image.licenses=Apache-2.0'
+      - '--label=org.opencontainers.image.name=prober'
```
yolocs, line 323:
```suggestion
      - '--label=org.opencontainers.image.name=jvs-prober'
```

.goreleaser.docker.yaml
```diff
@@ -280,6 +303,52 @@ dockers:
...
+      - '--label=org.opencontainers.image.revision={{ .FullCommit }}'
+      - '--label=org.opencontainers.image.source={{ .GitURL }}'
+      - '--label=org.opencontainers.image.title=prober'
```
yolocs, line 326:
```suggestion
      - '--label=org.opencontainers.image.title=jvs-prober'
```

.goreleaser.docker.yaml
```diff
@@ -280,6 +303,52 @@ dockers:
...
+    dockerfile: ./prober/prober.dockerfile
+    extra_files:
+    - ./prober/prober.sh
+    build_flag_templates:
+      - '--platform=linux/amd64'
+      - '--pull'
+      - '--label=org.opencontainers.image.created={{ .CommitTimestamp }}'
+      - '--label=org.opencontainers.image.description=Prober service for jvs'
+      - '--label=org.opencontainers.image.licenses=Apache-2.0'
+      - '--label=org.opencontainers.image.name=prober'
```
yolocs, line 346:
```suggestion
      - '--label=org.opencontainers.image.name=jvs-prober'
```

prober/prober.sh
capri-xiyue: I thought it still makes sense to throw error logs here so that when debug, at least we have some specific error messages we know to search.

.goreleaser.docker.yaml
```diff
@@ -280,6 +303,52 @@ dockers:
...
+      - '--label=org.opencontainers.image.revision={{ .FullCommit }}'
+      - '--label=org.opencontainers.image.source={{ .GitURL }}'
+      - '--label=org.opencontainers.image.title=prober'
```
yolocs, line 349:
```suggestion
      - '--label=org.opencontainers.image.title=jvs-prober'
```

prober/prober.sh
yolocs, line 11: Breakglass is not necessary because it won't contact server if it's breakglass.

yolocs (changes requested):

prober/prober.sh
capri-xiyue: Is it useful to add breakglass here so that we know whether the breakglass CUJ works or not?

yolocs: @capri-xiyue Not sure I understand your example. When the CLI can't validate the given token, it's an error https://github.com/abcxyz/jvs/blob/main/pkg/cli/validate.go#L166-L169

prober/prober.sh
yolocs: Is the error message here sufficient? https://github.com/abcxyz/jvs/blob/main/pkg/cli/validate.go#L166-L169

prober/prober.sh
yolocs: We don't need breakglass. When it's breakglass, CLI won't contact server. Remember this is _prober_, not unit test of the CLI

prober/prober.sh
capri-xiyue: I thought we will test both breakglass and nonbreakglass cases. Therefore, just the error messages from the CLI still looks confusing because I need to figure out the error message is from which CUJ. Since we won't test breakglass cases here, I don't think we need to add other error messages. If the job fail, I know that's for nonbreakglass case.

prober/prober.sh
capri-xiyue: Yep, I double checked the code, it will throw error instead of saying it's a success just can't validate.

prober/prober.sh
yolocs: Right, breakglass case should already be covered in unit test because it's self-contained.

prober/prober.sh
capri-xiyue, line 9: jvs_prober?

prober/prober.sh
capri-xiyue, line 5: jvs_prober?

prober/prober.dockerfile
sailorlqh: It will have a permission denied error whens starting the container: `terminated: Application failed to start: invalid status ::14: could not start container: permission denied`.

I think we don't want to run `chmod` command on the script to solve this problem. So I chose to use `ENTRYPOINT ["bash", "prober.sh"]`

yolocs: > I think we don't want to run chmod command on the script to solve this problem

Even if it's not this problem, we should `chmod +x` to make the script executable. If it solves the problem, even better.

prober/prober.sh
sailorlqh: Done,

prober/prober.sh
sailorlqh: Done.

prober/prober.dockerfile
sailorlqh: Done.

prober/prober.dockerfile
sailorlqh: Done.

prober/prober.sh
sailorlqh: Talked with Seth, made sure it's actually '-sf'. Fixed now.

capri-xiyue (approved):

.goreleaser.docker.yaml
```diff
@@ -280,6 +303,52 @@ dockers:
       - '--label=org.opencontainers.image.title=jvs-ui'
       - '--label=org.opencontainers.image.version={{ .Version }}'
 
+  -
+    ids:
+    - jvsctl
```
yolocs, line 308: Sorry, it just occurred to me this should be named as `jvs-prober`. We used the `jvsctl` binary but the container image is a prober.

.goreleaser.docker.yaml
```diff
@@ -280,6 +303,52 @@ dockers:
...
+      - '--label=org.opencontainers.image.licenses=Apache-2.0'
+      - '--label=org.opencontainers.image.name=jvs-prober'
+      - '--label=org.opencontainers.image.revision={{ .FullCommit }}'
+      - '--label=org.opencontainers.image.source={{ .GitURL }}'
+      - '--label=org.opencontainers.image.title=jvs-prober'
+      - '--label=org.opencontainers.image.version={{ .Version }}'
+
+  -
+    ids:
+    - jvsctl
```
yolocs, line 331: Same here.

.goreleaser.docker.yaml
```diff
@@ -280,6 +303,52 @@ dockers:
...
       - '--label=org.opencontainers.image.version={{ .Version }}'
 
+  -
+    ids:
+    - jvsctl
+    use: 'buildx'
+    goos: 'linux'
+    goarch: 'amd64'
+    image_templates:
+      - '{{ .Env.CONTAINER_REGISTRY }}/prober:{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}-amd64'
```
yolocs, line 313:
```suggestion
      - '{{ .Env.CONTAINER_REGISTRY }}/jvs-prober:{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}-amd64'
```

.goreleaser.docker.yaml
```diff
@@ -280,6 +303,52 @@ dockers:
...
+      - '--label=org.opencontainers.image.version={{ .Version }}'
+
+  -
+    ids:
+    - jvsctl
+    use: 'buildx'
+    goos: 'linux'
+    goarch: 'arm64'
+    image_templates:
+      - '{{ .Env.CONTAINER_REGISTRY }}/prober:{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}-arm64'
```
yolocs, line 336: Same here

.goreleaser.docker.yaml
```diff
@@ -301,6 +370,11 @@ docker_manifests:
     image_templates:
       - '{{ .Env.CONTAINER_REGISTRY }}/jvs-ui:{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}-amd64'
       - '{{ .Env.CONTAINER_REGISTRY }}/jvs-ui:{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}-arm64'
+  -
+    name_template: '{{ .Env.CONTAINER_REGISTRY }}/prober:{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}'
+    image_templates:
+      - '{{ .Env.CONTAINER_REGISTRY }}/prober:{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}-amd64'
+      - '{{ .Env.CONTAINER_REGISTRY }}/prober:{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}-arm64'
```
yolocs, line 377: Same here

prober/prober.dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+# Normally we would set this to run as "nobody".
+# But goreleaser builds the binary locally and sometimes it will mess up the permission
+# and cause "exec user process caused: permission denied".
+#
+# USER nobody
+
+# Run the bash script on container startup.
+
+RUN chmod +x /prober.sh
```
yolocs, line 15: This is fine. But I believe if you run the command locally, the script will stay an executable when copied into the container.

yolocs (changes requested):

yolocs (approved):

PR #242, opened 2023-04-07
(wip)add real integration test for api path

sqin2019:
Two issues
1. KMS key must have primary, current [key](https://github.com/abcxyz/jvs/blob/7ee0094e7aff40d14915b6ab8975b2198d5e3117/terraform/modules/jvs-services/key.tf#L15) does not have primary version set upon creation, or it must get rotated before justification/validation integ tests.
2. Need two ID token for authentication since aud (cloud run URL) is different, but [google-github-actions/auth](https://github.com/google-github-actions/auth) seems can only have one in a single job? Need more investigation. --> fixed, typo issue.


test/integration/main_test.go
```diff
@@ -0,0 +1,167 @@
+// Copyright 2023 The Authors (see AUTHORS file)
```
sethvargo, line 1: There is already a [test/integ](https://github.com/abcxyz/jvs/tree/main/test/integ) folder. We might want to rethink how we are structuring these.

sqin2019: I am thinking we only need the test/integration, the test/integ is removed. What is your suggestion?

sethvargo: Hey @sqin2019 - sorry, I missed this comment in my notifications 😦 

Taking a step back: what exactly are we trying to integration test? If https://github.com/abcxyz/jvs/pull/245 is merged, there's a shared framework for starting all the services.

Should we move the integration tests closers to the servers? The tests in #245 already do a basic health check, and they could be expanded to do more if needed.

sqin2019: Hi @sethvargo no worries! 

For the api path, we can use jvsctl to call each API and verify the output, will use the new framework for this.

For moving the integration tests closer to the servers, do you mean to split them up and move them to the corresponding server packages? 

For expanding the health check, do you mean to expand and add it in the integration tests? If so, it can be a follow up once integration tests is ready.

sethvargo: > For moving the integration tests closer to the servers, do you mean to split them up and move them to the corresponding server packages?

Yes

> For expanding the health check, do you mean to expand and add it in the integration tests? If so, it can be a follow up once integration tests is ready.

Let's table this for a longer discussion.

test/integration/config.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/sethvargo/go-envconfig"
+)
+
+type config struct {
+	APIAuthToken         string `env:"INTEG_TEST_CERT_TOTATOER_AUTH_TOKEN,required"`
```
verbanicm, line 26: This appears to be a copy/paste issue

PR #245, opened 2023-04-11
Switch to using a single binary

sethvargo:
This is a pretty significant rearchitecture that moves away from publishing one-binary-per-service to publishing a full binary that contains the CLI and server components. It is capable of running CLI commands, servers, or both. This dramatically reduces the number of binaries and container images that we need to manage, and it reduces the cognitive complexity for customers who have to choose between the various binaries to download.

This PR will almost certainly be imperfect, especially with the intersection of Terraform and probers. We can iterate, but I think this should be mergeable in a backwards-compatible way.


Dockerfile
```diff
@@ -1,22 +1,12 @@
-# Use distroless for ca certs.
-FROM gcr.io/distroless/static AS distroless
+FROM gcr.io/distroless/static
```
sethvargo, line 1: I just decided to use distroless/static directly, since the binary is pretty fat and might need the other things distroless provides.

cmd/cert-rotation/main.go
```diff
@@ -17,20 +17,12 @@ package main
 import (
 	"context"
 	"fmt"
-	"net/http"
```
sethvargo, line 20: For these commands, their business logic has been moved into pkg/cli almost copy-paste. I'll call out specific things inline where I made changes.

The new binaries are just stubs that "exec" the CLI. We could opt to delete these entirely instead.

docs/web-ui.md
```diff
@@ -20,32 +20,32 @@ DEV_MODE="true"
 
 ```shell
 ## A semi-colon separated string denoting the allowed domains and/or subdomains. This field is required.
-ALLOWLIST="example.com;foo.bar.com"
+ALLOWLIST="example.com, foo.bar.com"
```
sethvargo, line 23: This is an unrelated change, but it was bothering me. I don't understand why we changed the default delimeter. @shan4993 do you remember?

pkg/cli/api_server.go
```diff
@@ -0,0 +1,120 @@
...
+type APIServerCommand struct {
+	cli.BaseCommand
+
+	// testLookuper overrides the lookuper. It is only used for testing.
+	testLookuper envconfig.Lookuper
+
+	// testBoundAddress is populated by the listener. It is only used for testing
+	// where the server is configured to listen on a raondom port, but tests need
+	// to know the port and address it chose.
+	testBoundAddress string
```
sethvargo, line 49: These new pkg/cli files are almost identical to the `realMain` bodies from `cmd` with this exception:

- There's an ability to inject custom environments into the command during parsing (which propagates to envconfig).
- There's a private field which exposes the bind address. When services bind to port "0", the operating system selects a random, unused port. We use this to our advantage for parallel testing, where we can fully parallelize all tests by allowing the OS to pick a port, but we need to push the port back up the stack in the test so that clients can connect.

pkg/cli/api_server.go
```diff
@@ -0,0 +1,120 @@
...
+
+	grpcServer := grpc.NewServer(grpc.ChainUnaryInterceptor(
+		logging.GRPCUnaryInterceptor(logger, cfg.ProjectID),
+		otelgrpc.UnaryServerInterceptor(),
+	))
+
+	// Create basic health check
+	hs := health.NewServer()
+	hs.SetServingStatus("", healthpb.HealthCheckResponse_SERVING)
+	healthpb.RegisterHealthServer(grpcServer, hs)
```
sethvargo, line 98: This is a change you'll see throughout. All endpoints expose a super dumb health check, and the tests call the health check and assert a happy response. This is just for CLI testing and is not meant to replace a real integration test.

pkg/cli/root.go
```diff
@@ -33,9 +35,53 @@ var rootCmd = func() cli.Command {
 		Name:    "jvsctl",
 		Version: version.HumanVersion,
 		Commands: map[string]cli.CommandFactory{
+			"api": func() cli.Command {
```
sethvargo, line 38: I'm open to other names instead of subcommands. `ui-server` and `public-key-server` and `cert-rotation-server` felt bad. I figured we might want `jvsctl rotation invoke` in the future for manual rotation or something, so I elected to do namespaced subcommands for everything.

pkg/config/cert_rotation_config.go
```diff
@@ -21,6 +21,9 @@ import "github.com/hashicorp/go-multierror"
 type CertRotationConfig struct {
 	*CryptoConfig
 
+	// ProjectID is the Google Cloud project ID.
```
sethvargo, line 24: This is a change throughout. I'm not sure where to note it, but this requires services to set a `PROJECT_ID` envvar instead of doing the gcpmetadata lookup. I _think_ I need to update the Terraform to inject these, but I wanted to get feedback before doing all that work.

Long story short, I deleted the wrong KMS keys because I had `gcloud` logged into a different project (which the gcpmetadata package picks up). I'd prefer if we fully owned the config.

terraform/modules/jvs-services/cert-rotator.tf
```diff
@@ -26,13 +26,17 @@ module "cert_rotator_cloud_run" {
 
   region = var.region
   name   = "jvs-cert-rotator"
-  image  = var.jvs_cert_rotator_service_image
+  image  = var.jvs_container_image
+  args   = ["rotation", "server"]
```
sethvargo, line 30: Since everything is the same binary, we rely on args to spawn the correct process.

terraform/modules/jvs-services/cert-rotator.tf
```diff
@@ -26,13 +26,17 @@ module "cert_rotator_cloud_run" {
...
 
   # Cert rotator is not a user facing service. Ignore the ingress input.
   ingress = "all"
 
   service_account_email = var.jvs_cert_rotator_service_account
-  envvars               = merge({ "KEY_NAMES" : google_kms_crypto_key.signing_key.id }, var.cert_rotator_envvars)
+
+  envvars = merge({
+    "KEY_NAMES" : google_kms_crypto_key.signing_key.id
```
sethvargo, line 38: TODO(sethvargo): inject PROJECT_ID in here unless @yolocs tells me it's a terrible idea :)

.goreleaser.docker.yaml
```diff
@@ -351,25 +136,11 @@ dockers:
...
-      - '{{ .Env.CONTAINER_REGISTRY }}/jvs-public-key:{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}-amd64'
-      - '{{ .Env.CONTAINER_REGISTRY }}/jvs-public-key:{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}-arm64'
-  -
-    name_template: '{{ .Env.CONTAINER_REGISTRY }}/jvs-ui:{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}'
-    image_templates:
-      - '{{ .Env.CONTAINER_REGISTRY }}/jvs-ui:{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}-amd64'
-      - '{{ .Env.CONTAINER_REGISTRY }}/jvs-ui:{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}-arm64'
+      - '{{ .Env.CONTAINER_REGISTRY }}/jvsctl:{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}-amd64'
+      - '{{ .Env.CONTAINER_REGISTRY }}/jvsctl:{{ if index .Env "DOCKER_TAG"  }}{{ .Env.DOCKER_TAG }}{{ else }}{{ .Version }}{{ end }}-arm64'
+
```
sethvargo, line 143: This diff is really difficult to parse; I recommend just looking at the file. We build 1 binary (jvsctl) and publish 2 containers (jvsctl and jvs-prober). We have to publish the prober container separately to ensure bash exists and to inject the prober script.

verbanicm:
LGTM from a terraform and actions perspective

Dockerfile
yolocs: Would that be the recommendation going forward? cgr.dev/chainguard/static:latest chainguard alternative claims zero CVEs

pkg/cli/api_server.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/abcxyz/pkg/cli"
+	"github.com/abcxyz/pkg/logging"
+	"github.com/abcxyz/pkg/serving"
+	"github.com/sethvargo/go-envconfig"
+	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
+	"google.golang.org/api/option"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/health"
+	healthpb "google.golang.org/grpc/health/grpc_health_v1"
+	"google.golang.org/grpc/reflection"
```
yolocs, line 36: nit: https://github.com/abcxyz/readability/wiki/Go-Best-Practices#order--group-imports

Could be a follow up. Due to the change size, it's probably a better idea to get this in first.

pkg/cli/api_server.go
```diff
@@ -0,0 +1,132 @@
...
+	if err != nil {
+		return err
+	}
+	defer closer()
+
+	return server.StartGRPC(ctx, grpcServer)
+}
+
+func (c *APIServerCommand) RunUnstarted(ctx context.Context, args []string) (*serving.Server, *grpc.Server, func(), error) {
+	closer := func() {}
```
yolocs, line 79: Why this?

pkg/cli/api_server_test.go
```diff
@@ -0,0 +1,122 @@
...
+				envconfig.MapLookuper(map[string]string{
+					// Make the test choose a random port.
+					"PORT": "0",
+				}),
+			)
+			cmd.testKMSClientOptions = []option.ClientOption{
+				// Disable auth lookup in these tests, since we don't actually call KMS.
+				option.WithoutAuthentication(),
+			}
+			_, _, _ = cmd.Pipe()
```
yolocs, line 85: Why this?

pkg/cli/root.go
yolocs: The existing names in the PR are fine to me. One nit: should they be `jvsctl api serve` rather than `server`?

Unrelated, I wonder what about the existing CLI names. Should they be

`jvsctl token create` and `jvsctl token validate`

pkg/config/cert_rotation_config.go
yolocs: I think we've discussed this before. `PROJECT_ID` should be a universal env var for all of our services.

yolocs (approved):

PR #248, opened 2023-04-12
cleanup: Use FlagSet for configuration

yolocs:
Massive clean up:

- Used `cli.FlagSet` for all configurations; updated config unit tests to test with `cli.FlagSet`
- Added prefix to all server env vars
- Reenforced config validation to check `PROJECT_ID` and keys because they are required
- Provided defaults for all defaultable configs
- Merged `crypto_config.go` into `cert_rotation_config.go` because it's not used anywhere else
- Removed all YAML support because it was not used
- Removed useless version config
- Embedded `JustificationConfig` in `UIServiceConfig` for easier config loading

Sample helper output:

```
jvsctl ui server -h
Usage: jvsctl ui server [options]

  Start a UI server for the JVS.

COMMON SERVER OPTIONS

    -dev
        Set to true to enable more granular debugging in logs The default
        value is "false". This option can also be specified with the DEV_MODE
        environment variable.

    -port="string"
        The port the server listens to. The default value is "8080". This option
        can also be specified with the PORT environment variable.

    -project-id="string"
        Google Cloud project ID. This option can also be specified with the
        PROJECT_ID environment variable.

API OPTIONS

    -default-ttl="time.Duration"
        The default TTL for JVS tokens if not specified in the request. The
        default value is "15m". This option can also be specified with the
        JVS_API_DEFAULT_TTL environment variable.

    -issuer="string"
        The value to set to the issuer claim when signing JVS tokens. The
        default value is "jvs.abcxyz.dev". This option can also be specified
        with the JVS_API_ISSUER environment variable.

    -key-name="projects/[JVS_PROJECT]/locations/global/keyRings/[JVS_KEYRING]/cryptoKeys/[JVS_KEY]"
        The KMS key for signing JVS tokens. This option can also be specified
        with the JVS_KEY environment variable.

    -max-ttl="time.Duration"
        The maximum TTL that a token can have. The default value is "4h".
        This option can also be specified with the JVS_API_MAX_TTL environment
        variable.

    -signer-cache-timeout="time.Duration"
        The duration that keys stay in cache before being revoked. The
        default value is "5m". This option can also be specified with the
        JVS_API_SIGNER_CACHE_TIMEOUT environment variable.

UI OPTIONS

    -allowlist="example.com,*.foo.bar"
        List of allowed domains. This option can also be specified with the
        JVS_UI_ALLOWLIST environment variable.
```


pkg/config/cert_rotation_config.go
```diff
@@ -29,15 +33,136 @@ type CertRotationConfig struct {
...
 
 // Validate checks if the config is valid.
-func (c *CertRotationConfig) Validate() error {
-	var merr *multierror.Error
+func (cfg *CertRotationConfig) Validate() error {
+	var err *multierror.Error
+
+	if cfg.KeyTTL <= 0 {
+		err = multierror.Append(err, fmt.Errorf("key ttl must be a positive duration, got %q", cfg.KeyTTL))
+	}
```
sethvargo, line 63: Nit, prefer `got` shadow for easier copy-paste:

```suggestion
	if got := cfg.KeyTTL; got <= 0 {
		err = multierror.Append(err, fmt.Errorf("key ttl must be a positive duration, got %q", got))
	}
```

pkg/config/cert_rotation_config.go
```diff
@@ -29,15 +33,136 @@ type CertRotationConfig struct {
...
+		Default: "8080",
+		Usage:   `The port the server listens to.`,
+	})
+
+	f = set.NewSection("KEY ROTATION OPTIONS")
+
+	f.DurationVar(&cli.DurationVar{
+		Name:    "key-ttl",
+		Target:  &cfg.KeyTTL,
+		EnvVar:  "KEY_TTL",
```
sethvargo, line 130: With the exception of port, dev, and project_id, should some of these envvars be prefixed?

sethvargo (approved):

PR #249, opened 2023-04-12
fix: refactor caused prober issue

sailorlqh:


prober/prober.dockerfile
```diff
@@ -4,7 +4,6 @@ FROM cgr.dev/chainguard/bash:latest
 COPY jvsctl /jvsctl
 
 COPY ./prober/prober.sh /prober.sh
-RUN chmod +x /prober.sh
```
sethvargo, line 7: Do we no longer need to mark the prober script as executable?

sailorlqh: We will make the file executable locally, and when we copy it into container it stays executable.

PR #250, opened 2023-04-12
Feat: Update e2e code to include monitoring and prober service

sailorlqh:
There will be another PR for the infra repo to deploy jvs with monitoring and prober into our jvs-dev env.

### DIFF
--- a/terraform/e2e/main.tf
+++ b/terraform/e2e/main.tf
@@ -67,3 +67,19 @@ module "jvs_services" {
   kms_keyring_id = module.jvs_common.kms_keyring_id
   kms_key_name   = "signing-${random_id.default.hex}"
 }
+
+module "jvs_monitoring" {
+  source = "../modules/monitoring"
+
+  project_id = var.project_id
+
+  jvs_service_name               = module.jvs_services.jvs_api_service_name
+  cert_rotator_service_name      = module.jvs_services.jvs_cert_rotator_service_name
+  public_key_service_name        = module.jvs_services.jvs_public_key_service_name
+  jvs_ui_service_name            = module.jvs_services.jvs_ui_service_name
+  notification_channel_email     = var.notification_channel_email
+  prober_jvs_api_address         = join(":", [substr(module.jvs_services.jvs_api_service_url, 8, -1), "443"])
+  prober_jvs_public_key_endpoint = join("/", [module.jvs_services.jvs_public_key_service_url, ".well-known", "jwks"])
+  jvs_prober_image               = var.jvs_prober_image
+  prober_audience                = var.prober_audience
+}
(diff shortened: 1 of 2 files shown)


capri-xiyue (approved):

terraform/e2e/variables.tf
```diff
@@ -54,3 +54,18 @@ variable "jvs_container_image" {
...
+}
+
+variable "jvs_prober_image" {
+  type        = string
+  description = "docker image for jvs-prober"
+}
+
+variable "prober_audience" {
+  type        = string
+  description = "The cloud run url for jvs api service or app address."
```
yolocs, line 70: Should we default to the gcloud app?

terraform/e2e/main.tf
```diff
@@ -67,3 +67,19 @@ module "jvs_services" {
...
+  source = "../modules/monitoring"
+
+  project_id = var.project_id
+
+  jvs_service_name               = module.jvs_services.jvs_api_service_name
+  cert_rotator_service_name      = module.jvs_services.jvs_cert_rotator_service_name
+  public_key_service_name        = module.jvs_services.jvs_public_key_service_name
+  jvs_ui_service_name            = module.jvs_services.jvs_ui_service_name
+  notification_channel_email     = var.notification_channel_email
+  prober_jvs_api_address         = join(":", [substr(module.jvs_services.jvs_api_service_url, 8, -1), "443"])
```
yolocs, line 81: Should we use https://developer.hashicorp.com/terraform/language/functions/trimprefix?

yolocs: Also please add a comment to explain.

yolocs: `= "${trimprefix(module.jvs_services.jvs_api_service_url, \"https://\")}:443"` - would this work?

terraform/e2e/main.tf
```diff
@@ -67,3 +67,19 @@ module "jvs_services" {
...
+  prober_jvs_public_key_endpoint = join("/", [module.jvs_services.jvs_public_key_service_url, ".well-known", "jwks"])
```
yolocs, line 82:
```suggestion
  prober_jvs_public_key_endpoint = "${module.jvs_services.jvs_public_key_service_url}/.well-known/jwks"
```

yolocs, line 82: On a 2nd thought, I think these should be input variables. We should use the real domain to do the probing instead of directly probing the cloud run services.

terraform/e2e/variables.tf
sailorlqh: Yes, my idea is in the infra repo, we will put the gcloud app url.

terraform/e2e/main.tf
sailorlqh: If we are going to request through gclb, we can just use jvs_api_domin here. I have updated the code.

terraform/e2e/main.tf
sailorlqh: Since we are going to request through gclb, we can just use jvs_api_domin here instead of using cloud run service url.

sailorlqh: Already updated to use domain address.

yolocs (approved):

PR #256, opened 2023-04-14
fix: use service identity resource to retrieve IAP SA

yolocs:


sethvargo (approved):
Nit: this doesn't "retrieve" the IAP SA, it creates the P4SA.

PR #262, opened 2023-04-19
fix: combine enable switch for basic monitoring and prober monitoring

sailorlqh:
fix issues/261.

### DIFF
--- a/docs/prober-and-monitoring.md
+++ b/docs/prober-and-monitoring.md
@@ -46,7 +46,7 @@ module "jvs_monitoring" {
 }
 ```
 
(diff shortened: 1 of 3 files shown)


docs/prober-and-monitoring.md
```diff
@@ -46,10 +46,9 @@ module "jvs_monitoring" {
 }
 ```
 
-By default, alering is disabled, you can enable it by setting the following variables:
+By default, alering is disabled, you can enable alerting for cloud run services and prober by setting the following variables:
```
yolocs, line 49:
```suggestion
By default, alering is disabled, you can enable it by setting the following variables:
```

yolocs (approved):

PR #263, opened 2023-04-20
use healthcheck from pkg

shan4993:


pkg/cli/public_key_server_test.go
```diff
@@ -110,6 +110,8 @@ func TestPublicKeyServerCommand(t *testing.T) {
 				t.Fatal(err)
 			}
 
+			req.Header.Set("Accept", "application/json")
+
```
shan4993, line 114: maybe not needed?

yolocs: The test doesn't check body content; only status code. Should be ok to remove.

yolocs (approved):
Thanks, Shan!

Issue #246, opened 2023-04-11
Move some jvsctl commands to "token" subcommand


sethvargo:
I wonder what about the existing CLI names. Should they be

`jvsctl token create` and `jvsctl token validate`

_Originally posted by @yolocs in https://github.com/abcxyz/jvs/pull/245#discussion_r1163238036_

PR #266, opened 2023-04-24
Fix: Move some jvsctl commands to token subcommand

dandans-verily:
Move create and validate to token subcommand.
Fix: #246

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -54,5 +54,5 @@ If you installed JVS using the provided Terraform module as described in the
 authentication. E.g.
 
 ```sh
-jvsctl token --auth-token $(gcloud auth print-identity-token) -e "just testing"
+jvsctl token validate --auth-token $(gcloud auth print-identity-token) -e "just testing"
 ```
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -54,5 +54,5 @@ If you installed JVS using the provided Terraform module as described in the
 authentication. E.g.
 
 ```sh
-jvsctl token validate --auth-token $(gcloud auth print-identity-token) -e "just testing"
+jvsctl token create --auth-token $(gcloud auth print-identity-token) -e "just testing"
 ```
--- a/pkg/cli/root.go
+++ b/pkg/cli/root.go
@@ -69,7 +69,18 @@ var rootCmd = func() cli.Command {
 				}
 			},
 			"token": func() cli.Command {
-				return &TokenCommand{}
+				return &cli.RootCommand{
+					Name:        "token",
+					Description: "Perform token operations",
+					Commands: map[string]cli.CommandFactory{
+						"create": func() cli.Command {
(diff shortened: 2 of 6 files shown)


sethvargo (approved):

docs/cli.md
```diff
@@ -54,5 +54,5 @@ If you installed JVS using the provided Terraform module as described in the
 authentication. E.g.
 
 ```sh
-jvsctl token --auth-token $(gcloud auth print-identity-token) -e "just testing"
+jvsctl token validate --auth-token $(gcloud auth print-identity-token) -e "just testing"
```
yolocs, line 57: This _was_ a token create command.
```suggestion
jvsctl token create --auth-token $(gcloud auth print-identity-token) -e "just testing"
```

pkg/cli/root_test.go
```diff
@@ -28,9 +28,8 @@ Usage: jvsctl COMMAND
   api           Perform API operations
   public-key    Perform public-key operations
   rotation      Perform rotation operations
-  token         Generate a justification token
+  token         Perform token operations
```
yolocs, line 31: Can you also update https://github.com/abcxyz/jvs/blob/main/prober/prober.sh to reflect the change?

yolocs (changes requested):

pkg/cli/root_test.go
sqin2019: maybe https://github.com/abcxyz/jvs/blob/main/docs/cli.md#authentication as well?

dandans-verily: Updated https://github.com/abcxyz/jvs/blob/main/prober/prober.sh in a new commit. PTAL :D

pkg/cli/root.go
```diff
@@ -69,7 +69,18 @@ var rootCmd = func() cli.Command {
 				}
 			},
 			"token": func() cli.Command {
-				return &TokenCommand{}
+				return &cli.RootCommand{
+					Name:        "token",
+					Description: "Perform token operations",
+					Commands: map[string]cli.CommandFactory{
+						"create": func() cli.Command {
+							return &TokenCommand{}
```
yolocs, line 77: nit: Can we now rename this struct to `TokenCreateCommand`?

pkg/cli/root.go
```diff
@@ -69,7 +69,18 @@ var rootCmd = func() cli.Command {
...
+						},
+						"validate": func() cli.Command {
+							return &ValidateCommand{}
```
yolocs, line 80: nit: Can we now rename this struct to `TokenValidateCommand`?

yolocs (approved):

pkg/cli/root.go
dandans-verily: Yes, I will rename in a separate followed-up PR so it is easier to see all the other relevant changes.

Issue #258, opened 2023-04-14
Deploy new prober in the release workflow


yolocs:
https://github.com/abcxyz/jvs/blob/main/.github/workflows/release.yml - should also update the prober in the dev env

PR #269, opened 2023-05-08
Deploy new prober in the release workflow

dandans-verily:
Fix: #258

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -116,7 +116,7 @@ jobs:
       with:
         workload_identity_provider: '${{ env.WIF_PROVIDER }}'
         service_account: '${{ env.WIF_SERVICE_ACCOUNT }}'
-    - name: 'Deploy Cloud Run services'
+    - name: 'Deploy Cloud Run services and update prober job'
       run: |-
         # Image tags are without 'v' prefix.
         DOCKER_TAG="${{ github.ref_name }}"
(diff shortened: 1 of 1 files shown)


sailorlqh (approved):

.github/workflows/release.yml
```diff
@@ -126,3 +126,9 @@ jobs:
           --project="${{ env.DEV_PROJECT_ID }}" \
           --region="${{ env.DEV_REGION }}" \
           --image="${{ env.CONTAINER_REGISTRY }}/jvsctl:${DOCKER_TAG}-amd64"
+
+        # Update Prober job.
+        gcloud run jobs update jvs-prober \
```
sethvargo, line 131: I don't know if this is correct. This is a _matrix_ job, so it runs once per each matrix entry defined on L112. This will update the prober 4x, which is not what we want.

dandans-verily: The items inside the matrix are all services deployed via [1], and need to be updated via "gcloud run services update", while prober is a job deployed via [2] and need to be updated by "gcloud run jobs update". Let me know whether that is how things are deployed in dev project tycho-jvs-dev-e06f, happy to update accordingly :D

[1] https://github.com/abcxyz/jvs/blob/main/terraform/e2e/main.tf#L57
[2] https://github.com/abcxyz/jvs/blob/main/terraform/modules/monitoring/prober.tf#L17

yolocs (changes requested):

.github/workflows/release.yml
dandans-verily: Thanks for the details, good to learn about the matrix strategy :D Agree that it is better to split it to a separate job in this case.

Updated. PTAL :D

.github/workflows/release.yml
```diff
@@ -126,3 +126,28 @@ jobs:
           --project="${{ env.DEV_PROJECT_ID }}" \
           --region="${{ env.DEV_REGION }}" \
           --image="${{ env.CONTAINER_REGISTRY }}/jvsctl:${DOCKER_TAG}-amd64"
+
+
+  deploy-dev-job:
```
yolocs, line 131: nit
```suggestion
  deploy-dev-prober:
```

yolocs (approved):

sethvargo (approved):

PR #289, opened 2023-06-12
chore: use multicloser in server commands

yolocs:


pkg/cli/api_server.go
```diff
@@ -67,16 +68,20 @@ func (c *APIServerCommand) Flags() *cli.FlagSet {
...
 	}
-	defer closer()
 
 	return server.StartGRPC(ctx, grpcServer)
 }
 
-func (c *APIServerCommand) RunUnstarted(ctx context.Context, args []string) (*serving.Server, *grpc.Server, func(), error) {
-	closer := func() {}
+func (c *APIServerCommand) RunUnstarted(ctx context.Context, args []string) (*serving.Server, *grpc.Server, *multicloser.Closer, error) {
+	closer := &multicloser.Closer{}
```
sethvargo, line 84: Nit: 

```suggestion
	var closer *multicloser.Closer
```

(throughout)

sethvargo (approved):

pkg/cli/api_server.go
yolocs: Wouldn't it be safer to do `closer := &multicloser.Closer{}`? Otherwise returning a `nil` closer and calling `closer.Close()` will panic. We don't have safe func to call to handle nil

yolocs: Maybe we do have safe func... https://github.com/abcxyz/pkg/blob/main/multicloser/multicloser.go#L73

sethvargo: I engineered it to be safe 😄

yolocs: Done

pkg/cli/api_server_test.go
```diff
@@ -84,10 +84,14 @@ func TestAPIServerCommand(t *testing.T) {
 			_, _, _ = cmd.Pipe()
 
 			srv, grpcServer, closer, err := cmd.RunUnstarted(ctx, tc.args)
+			defer func() {
+				if err := closer.Close(); err != nil {
+					t.Log(err)
```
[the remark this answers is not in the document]
yolocs: Done

sqin2019 (approved):

PR #290, opened 2023-06-13
Add JVS plugin interfaces

dandans-verily:
Split out this PR per corresponding conversations from https://github.com/abcxyz/jvs/pull/288

Resolve: #272


apis/v0/plugin_interface.go
```diff
@@ -0,0 +1,91 @@
...
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/go-plugin"
+	grpc "google.golang.org/grpc"
+)
+
+const (
+	handshakeCookieKey   = "JVS_PLUGIN"
+	handshakeCookieValue = "hello"
```
sethvargo, line 27: Usually the cookie value is a long string. It's not meant for security, but "hello" is too generic. `openssl rand -hex 30` will give you a good value.

sethvargo (dismissed):

yolocs (dismissed):

apis/v0/plugin_interface.go
dandans-verily: Updated :D

yolocs (approved):

Issue #158, opened 2022-11-28
Use real services for integration test


yolocs:
At the moment, we don't run _real_ services in integration tests. Instead, we launch services in Go tests. This is not ideal since it doesn't provide test coverage on the real container images we release.

Per https://github.com/abcxyz/jvs/pull/157, we had a bug in the dockerfile that caused bad container images. The integration test didn't catch that and we only realized the issue when we were trying to deploy the images.

yolocs:
cc: @sqin2019

yolocs:
FYI: https://github.com/abcxyz/jvs/blob/main/.github/workflows/integration.yml doesn't really work.

PR #294, opened 2023-06-30
Load plugins when starting API server

dandans-verily:
Resolve: #277


pkg/cli/api_server.go
```diff
@@ -117,7 +118,14 @@ func (c *APIServerCommand) RunUnstarted(ctx context.Context, args []string) (*se
 	// Create basic health check
 	healthcheck.RegisterGRPCHealthCheck(grpcServer)
 
-	p := justification.NewProcessor(kmsClient, c.cfg)
+	validators, pluginClosers, err := plugin.LoadPlugins(c.cfg.PluginDir)
+	if err != nil {
+		logger.Debugw("error loading plugins", "error", err)
```
sethvargo, line 123: Wouldn't we want to fail and return an error in this case?

pkg/cli/api_server.go
```diff
@@ -117,7 +118,14 @@ func (c *APIServerCommand) RunUnstarted(ctx context.Context, args []string) (*se
...
+	}
+
+	closer = multicloser.Append(closer, pluginClosers.Close)
```
sethvargo, line 126:
```suggestion
	closer = multicloser.Append(closer, pluginClosers.Close)
```

pkg/justification/processor.go
```diff
@@ -148,6 +154,21 @@ func (p *Processor) runValidations(req *jvspb.CreateJustificationRequest) error
...
+			if e != nil {
+				err = multierror.Append(err, fmt.Errorf("unexpected error from validator %q: %w", j.Category, e))
+				continue
+			}
+
+			if !resp.Valid {
+				err = multierror.Append(err,
+					fmt.Errorf("failed validation criteria with error %v and warning %v", resp.Error, resp.Warning))
+			}
+		}
```
sethvargo, line 171: I don't think this is correct, since there's no logic for when the validator doesn't exist for the category. I think you might want:

```suggestion
		v, ok := p.validators[j.Category]
		if !ok {
			err = multierror.Append(err, fmt.Errorf("missing validator for category %q", j.Category))
		}
		
		resp, e := v.Validate(ctx, &jvspb.ValidateJustificationRequest{
			Justification: j,
		})
		if e != nil {
			err = multierror.Append(err, fmt.Errorf("unexpected error from validator %q: %w", j.Category, e))
			continue
		}

		if !resp.Valid {
			err = multierror.Append(err, fmt.Errorf("failed validation criteria with error %v and warning %v", resp.Error, resp.Warning))
		}
```

pkg/config/justification_config.go
```diff
@@ -45,6 +45,9 @@ type JustificationConfig struct {
 	// Issuer will be used to set the issuer field when signing JWTs
 	Issuer string `env:"JVS_API_ISSUER,overwrite,default=jvs.abcxyz.dev"`
 
+	// PluginDir is the directory to load plugins.
+	PluginDir string `env:"JVS_PLUGIN_DIR,overwrite,default=plugins"`
```
sethvargo, line 49: I would expect this to be an absolute path like `/var/jvs/plugins`

pkg/justification/processor_test.go
```diff
@@ -179,6 +190,88 @@ func TestCreateToken(t *testing.T) {
...
+				},
+				Ttl: durationpb.New(3600 * time.Second),
+			},
+			validators: map[string]jvspb.Validator{
+				"explanation": &mockValidator{
+					err: fmt.Errorf("Cannot connect to validator"),
+				},
+			},
+			wantErr: "unexpected error from validator \"explanation\": Cannot connect to validator",
+		},
```
sethvargo, line 274: Can you add a test where no validator exists for the category?

pkg/cli/api_server.go
dandans-verily: I was thinking maybe we still want to allow the host to be working by its own if plugins cannot be loaded. But since we are loading from the the plugin directory at deploy time, it makes sense to fail the server so we can manually trim the invalid plugins. Updated the code accordingly

pkg/cli/api_server.go
dandans-verily: Done

pkg/config/justification_config.go
dandans-verily: [plugin.Discover](https://github.com/hashicorp/go-plugin/blob/a88a423a8813d0b26c8e3219f71b0f30447b5d2e/discover.go#L19) can take both absolute and relative paths, so I was allowing similar things.

Updated this flag to take absolute path instead.

pkg/justification/processor.go
dandans-verily: Actually I am a bit confused by the `Category` handling logic in jvs repo, want to confirm the requirements below:

1. In the UI controller, `breakglass` is a category ([code pointer](https://github.com/abcxyz/jvs/blob/a02ba7d38f4b16c203d37e0d879b6c67883ec43a/pkg/controller/controller.go#L170)), but it will be immediately rejected [here](https://github.com/abcxyz/jvs/blob/a02ba7d38f4b16c203d37e0d879b6c67883ec43a/pkg/justification/processor.go#L148)
2. `explanation` is a category allowing free form text

I was thinking the additional plugins will add additional category on top of 1 and 2, and will be displayed as additional drop down items. As a result, `breakglass` and `explanation` would not have any corresponding validator plugins 

Let me know whether my understanding is correct, and also loop in @yolocs to this thread

pkg/config/justification_config.go
```diff
@@ -45,6 +46,9 @@ type JustificationConfig struct {
 	// Issuer will be used to set the issuer field when signing JWTs
 	Issuer string `env:"JVS_API_ISSUER,overwrite,default=jvs.abcxyz.dev"`
 
+	// PluginAbsDir is the absolute path of the directory to load plugins.
+	PluginAbsDir string `env:"JVS_PLUGIN_ABS_DIR,overwrite,default=/var/jvs/plugins"`
```
yolocs, line 50: Is there any particular reason to require abs dir?

pkg/justification/processor.go
yolocs: > In the UI controller, breakglass is a category (code pointer), but it will be immediately rejected here

This is actually a bug. I filed an issue https://github.com/abcxyz/jvs/issues/296. I will send a temporary fix for it.

For this PR, we can specially handle the "explanation" category. E.g.

1. Is "explanation"? Yes, generate token based on the request.
2. No, try find a validator. Found? Validate and generate token.
3. Not found. Fail.

pkg/justification/processor_test.go
dandans-verily: Added test case `missing_validator`

pkg/config/justification_config.go
dandans-verily: See related conversation from https://github.com/abcxyz/jvs/pull/294#discussion_r1247892221, my guess is it is easier to manage the dir structure of the image. Looping in @sethvargo for more information

pkg/config/justification_config.go
yolocs: I could be wrong but my understanding is that the **default** should be a path like `/vars/jvs/plugins`.

test/integ/main_test.go
```diff
@@ -91,6 +91,7 @@ func TestJVS(t *testing.T) {
 		ProjectID:          os.Getenv("PROJECT_ID"),
 		KeyName:            keyName,
 		Issuer:             "ci-test",
+		PluginAbsDir:       "/var/jvs/plugins",
```
dandans-verily, line 94: @yolocs I am not sure why this change is needed to make the integration test pass https://github.com/abcxyz/jvs/actions/runs/5479801335/jobs/9982133916. `go test` runs fine locally even without this change

Also it seems config.JustificationConfig is not loaded properly below? Otherwise it will pick up the default value

https://github.com/abcxyz/jvs/blob/1d813631a27efe248a1ea9675338d525dad93d78/test/integ/main_test.go#L90

pkg/config/justification_config.go
dandans-verily: You are looking at the stale version actually, that latest version has default value `/vars/jvs/plugins`. Here is the direct link: https://github.com/abcxyz/jvs/pull/294/files#diff-7968fc8c7a029e487f309044142bb04f5dda630cbb12256e474afb8f542cc015

yolocs: I meant we shouldn't require the config to be always absolute. `PluginDir string `env:"JVS_PLUGIN_DIR,overwrite,default=/vars/jvs/plugins"``

test/integ/main_test.go
yolocs: The short answer is: we don't load config in integ test. We directly construct a config in code (like the line you put).

The longer answer is: we should integ test more completely but didn't get time to. See https://github.com/abcxyz/jvs/issues/158

dandans-verily: I see, thanks for the context. Resolving this thread now

pkg/config/justification_config.go
dandans-verily: I don't have enough context to understand major difference between the two choices. @sethvargo and @yolocs could you provide more details on why or why not this should be absolute path?

yolocs: The config tells the main binary to seek plugins in the given dir, which could be anywhere in the system (absolute or not). But _by default_, that _should_ be a absolute path so the location of plugins won't change when you run the binary in different dirs.

pkg/justification/processor.go
```diff
@@ -68,14 +69,19 @@ const (
 	DefaultAudience = "dev.abcxyz.jvs"
 )
 
+func (p *Processor) WithValidators(v map[string]jvspb.Validator) *Processor {
```
yolocs, line 72: nit: comment

Also, thinking one step ahead. Maybe we want to expose a func to return the validators. In the `controller`, we need the validators to populate categories (and later hints, etc.)

yolocs (commented):
mostly nits

pkg/justification/processor.go
dandans-verily: Ah, yeah, plan to do that in the next PR when I am adjusting the controller to expose the category in UI. Will resolve this thread now

pkg/config/justification_config.go
dandans-verily: Thanks Chen for clarification. Updated the code to allow relative overwrite but set the absolute path as default.

pkg/config/justification_config.go
dandans-verily: Done

pkg/config/justification_config.go
```diff
@@ -66,6 +70,10 @@ func (cfg *JustificationConfig) Validate() error {
 		merr = multierror.Append(merr, fmt.Errorf("empty KeyName"))
 	}
 
+	if !filepath.IsAbs(cfg.PluginAbsDir) {
+		merr = multierror.Append(merr, fmt.Errorf("PluginAbsDir is not absolute path"))
+	}
+
```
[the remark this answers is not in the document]
dandans-verily: Done

pkg/justification/processor.go
```diff
@@ -146,7 +152,23 @@ func (p *Processor) runValidations(req *jvspb.CreateJustificationRequest) error
...
-			err = multierror.Append(err, fmt.Errorf("unexpected justification %v unrecognized", j))
+			v, ok := p.validators[j.Category]
+			if !ok {
+				err = multierror.Append(err, fmt.Errorf("missing validator for category %q", j.Category))
+				continue
+			}
+			resp, e := v.Validate(ctx, &jvspb.ValidateJustificationRequest{
+				Justification: j,
+			})
+			if e != nil {
```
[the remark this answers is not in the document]
dandans-verily: Done

pkg/justification/processor_test.go
```diff
@@ -179,6 +190,101 @@ func TestCreateToken(t *testing.T) {
 			},
 			wantErr: "must be less than 1000 bytes",
 		},
+		{
+			name: "happy_path_with_jira_validator",
```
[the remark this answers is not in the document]
dandans-verily: Done

yolocs (approved):

Issue #276, opened 2023-06-08
Adjust api_server to load the plugin


dandans-verily:

dandans-verily:
Done as part of https://github.com/abcxyz/jvs/pull/294

PR #298, opened 2023-07-08
Adjust a bunch of go-plugin implementation details to resolve plugin loading exceptions

dandans-verily:
Specifically fix the following three exceptions while loading the plugins

- plugin.ClientConfig.Plugins field need to be set otherwise exception will be thrown from https://github.com/hashicorp/go-plugin/blob/a88a423a8813d0b26c8e3219f71b0f30447b5d2e/client.go#L909
- Due to [type check](https://github.com/hashicorp/go-plugin/blob/a88a423a8813d0b26c8e3219f71b0f30447b5d2e/server.go#L191), ValidatorPlugin need to implement [plugin.GRPCPlugin](https://github.com/hashicorp/go-plugin/blob/a88a423a8813d0b26c8e3219f71b0f30447b5d2e/plugin.go#L36).
- Due to [dispense name check](https://github.com/hashicorp/go-plugin/blob/a88a423a8813d0b26c8e3219f71b0f30447b5d2e/grpc_client.go#L110), the name need to be consistent with the plugin name.

Link to: #276

### DIFF
--- a/apis/v0/plugin_interface.go
+++ b/apis/v0/plugin_interface.go
@@ -41,7 +41,9 @@ var Handshake = plugin.HandshakeConfig{
 
 // PluginMap is the map of plugins we can dispense.
 var PluginMap = map[string]plugin.Plugin{
-	"jvs-plugin-jira": &ValidatorPlugin{},
+	// Due to [dispense name check], the name need to be consistent with the plugin name.
+	// https://github.com/hashicorp/go-plugin/blob/a88a423a8813d0b26c8e3219f71b0f30447b5d2e/grpc_client.go#L110
+	"jira": &ValidatorPlugin{},
 }
 
 // The interface we are exposing as a plugin.
@@ -49,7 +51,9 @@ type Validator interface {
 	Validate(context.Context, *ValidateJustificationRequest) (*ValidateJustificationResponse, error)
 }
 
-// This is the implementation of plugin.GRPCPlugin so we can serve/consume this.
+// This is the implementation of [plugin.GRPCPlugin] so we can serve/consume this.
+//
+// [plugin.GRPCPlugin]: https://github.com/hashicorp/go-plugin/blob/a88a423a8813d0b26c8e3219f71b0f30447b5d2e/plugin.go#L36
 type ValidatorPlugin struct {
 	// GRPCPlugin must still implement the Plugin interface.
 	plugin.Plugin
@@ -58,12 +62,15 @@ type ValidatorPlugin struct {
 	Impl Validator
 }
 
-func (p *ValidatorPlugin) PluginServer(broker *plugin.GRPCBroker, s *grpc.Server) error {
+// Due to [type check], ValidatorPlugin need to implement the following interface.
+//
+// [type check]: https://github.com/hashicorp/go-plugin/blob/a88a423a8813d0b26c8e3219f71b0f30447b5d2e/server.go#L191
+func (p *ValidatorPlugin) GRPCServer(broker *plugin.GRPCBroker, s *grpc.Server) error {
 	RegisterJVSPluginServer(s, &PluginServer{Impl: p.Impl})
 	return nil
 }
 
-func (p *ValidatorPlugin) PluginClient(ctx context.Context, broker *plugin.GRPCBroker, c *grpc.ClientConn) (any, error) {
+func (p *ValidatorPlugin) GRPCClient(ctx context.Context, broker *plugin.GRPCBroker, c *grpc.ClientConn) (any, error) {
 	return &PluginClient{client: NewJVSPluginClient(c)}, nil
 }
 
--- a/apis/v0/plugin_interface.go
+++ b/apis/v0/plugin_interface.go
@@ -42,7 +42,8 @@ var Handshake = plugin.HandshakeConfig{
 // PluginMap is the map of plugins we can dispense.
 var PluginMap = map[string]plugin.Plugin{
 	// Due to [dispense name check], the name need to be consistent with the plugin name.
-	// https://github.com/hashicorp/go-plugin/blob/a88a423a8813d0b26c8e3219f71b0f30447b5d2e/grpc_client.go#L110
+	//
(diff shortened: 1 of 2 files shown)


apis/v0/plugin_interface.go
```diff
@@ -39,12 +39,22 @@ var Handshake = plugin.HandshakeConfig{
 	MagicCookieValue: handshakeCookieValue,
 }
 
+// PluginMap is the map of plugins we can dispense.
+var PluginMap = map[string]plugin.Plugin{
```
sethvargo, line 43: I'm not sure this should be public - we don't want anyone overridding the pluginmap outside of the package.

dandans-verily: Good point!!! I have moved it to plugin_manager and make it a package private variable now. Seems we need to adjust this variable every time a new plugin is support.

sethvargo (dismissed):

apis/v0/plugin_interface.go
```diff
@@ -44,7 +44,9 @@ type Validator interface {
 	Validate(context.Context, *ValidateJustificationRequest) (*ValidateJustificationResponse, error)
 }
 
-// This is the implementation of plugin.GRPCPlugin so we can serve/consume this.
+// This is the implementation of [plugin.GRPCPlugin] so we can serve/consume this.
```
yolocs, line 47:
```suggestion
// ValidatorPlugin implements [plugin.GRPCPlugin].
```

apis/v0/plugin_interface.go
```diff
@@ -53,12 +55,15 @@ type ValidatorPlugin struct {
 	Impl Validator
 }
 
-func (p *ValidatorPlugin) PluginServer(broker *plugin.GRPCBroker, s *grpc.Server) error {
+// Due to [type check], ValidatorPlugin need to implement the following interface.
```
yolocs, line 58:
```suggestion
// GRPCServer is required by [plugin.GRPCPlugin].
```

apis/v0/plugin_interface.go
```diff
@@ -53,12 +55,15 @@ type ValidatorPlugin struct {
...
+//
+// [type check]: https://github.com/hashicorp/go-plugin/blob/a88a423a8813d0b26c8e3219f71b0f30447b5d2e/server.go#L191
+func (p *ValidatorPlugin) GRPCServer(broker *plugin.GRPCBroker, s *grpc.Server) error {
 	RegisterJVSPluginServer(s, &PluginServer{Impl: p.Impl})
 	return nil
 }
 
-func (p *ValidatorPlugin) PluginClient(ctx context.Context, broker *plugin.GRPCBroker, c *grpc.ClientConn) (any, error) {
+func (p *ValidatorPlugin) GRPCClient(ctx context.Context, broker *plugin.GRPCBroker, c *grpc.ClientConn) (any, error) {
```
yolocs, line 66:
```suggestion
// GRPCClient is required by [plugin.GRPCPlugin].
func (p *ValidatorPlugin) GRPCClient(ctx context.Context, broker *plugin.GRPCBroker, c *grpc.ClientConn) (any, error) {
```

pkg/plugin/plugin_manager.go
```diff
@@ -51,7 +59,10 @@ func LoadPlugins(dir string) (map[string]jvspb.Validator, *multicloser.Closer, e
 
 		// Enable the plugin.
 		pluginClient := plugin.NewClient(&plugin.ClientConfig{
-			HandshakeConfig:  jvspb.Handshake,
+			HandshakeConfig: jvspb.Handshake,
+			// Plugins field need to be set otherwise exception will be thrown from
+			// https://github.com/hashicorp/go-plugin/blob/a88a423a8813d0b26c8e3219f71b0f30447b5d2e/client.go#L909
+			Plugins:          pluginMap,
```
yolocs, line 65: Why the `pluginMap` is static? Shouldn't it be based on the `name`? E.g.
```suggestion
			Plugins:          map[string]plugin.Plugin{name: &jvspb.ValidatorPlugin{}},
```

dandans-verily: My understanding is pluginMap demonstrates host's knowledge about each plugin as some validation check per the code pointer [here](https://github.com/hashicorp/go-plugin/blob/a88a423a8813d0b26c8e3219f71b0f30447b5d2e/examples/grpc/shared/interface.go#L26). If we make it a dynamic variable, we pretty much allows all the plugins in the dir. I have adjusted the code as suggest for now since it avoids the need to update the pluginMap every time a new plugin is introduced. @sethvargo let us know whether we use it the correct way.

pkg/plugin/plugin_manager.go
```diff
@@ -51,7 +51,15 @@ func LoadPlugins(dir string) (map[string]jvspb.Validator, *multicloser.Closer, e
...
 		// Enable the plugin.
 		pluginClient := plugin.NewClient(&plugin.ClientConfig{
-			HandshakeConfig:  jvspb.Handshake,
+			HandshakeConfig: jvspb.Handshake,
+			// Plugins field need to be set otherwise exception will be thrown from
+			// https://github.com/hashicorp/go-plugin/blob/a88a423a8813d0b26c8e3219f71b0f30447b5d2e/client.go#L909
+			Plugins: map[string]plugin.Plugin{
+				// Due to [dispense name check], the name need to be consistent with the plugin name.
+				//
+				// [dispense name check]: https://github.com/hashicorp/go-plugin/blob/a88a423a8813d0b26c8e3219f71b0f30447b5d2e/grpc_client.go#L110
```
yolocs, line 60: Due to the code context, we can probably save this comment.

yolocs (approved):

pkg/plugin/plugin_manager.go
dandans-verily: Done and add all the details to the PR description instead

yolocs (approved):

pkg/plugin/plugin_manager.go
dandans-verily: Resolving now to unblock the merge. @sethvargo feel free to keep commenting in this thread if any change is needed, I will send out follow up PR to make the change.

PR #300, opened 2023-07-11
adjust controller to use category from processor

dandans-verily:
Link to: #277


pkg/controller/controller.go
```diff
@@ -334,6 +343,36 @@ func ttls() []string {
...
+// getText returns the corresponding text for rendering.
+func getText(category string) string {
+	switch category {
+	case "explanation":
+		return "Explanation"
+	case "jira":
+		return "Jira"
+	default:
+		return "UNKNOWN"
+	}
```
sethvargo, line 377: This feels like it's going to be fragile and break over time. Can the plugin's expose this themselves instead?

pkg/justification/processor.go
```diff
@@ -69,11 +69,17 @@ const (
 	DefaultAudience = "dev.abcxyz.jvs"
 )
 
+// WithValidators add validators to the processor.
 func (p *Processor) WithValidators(v map[string]jvspb.Validator) *Processor {
 	p.validators = v
 	return p
 }
 
+// ValidatorMap returns all the validators allowed by this processor.
```
yolocs, line 78:
```suggestion
// Validators returns all the validators allowed by this processor.
```

pkg/controller/controller.go
```diff
@@ -27,20 +27,32 @@ import (
...
+
 	"golang.org/x/exp/slices"
 	"google.golang.org/protobuf/types/known/durationpb"
 )
 
+// processor is the mockable interface for the convenience of testing.
+type processor interface {
+	CreateToken(context.Context, string, *jvspb.CreateJustificationRequest) ([]byte, error)
+	Validators() map[string]jvspb.Validator
+}
```
yolocs, line 39: I think you can still use the existing processor? You just create a processor with validators.

pkg/controller/controller.go
```diff
@@ -252,10 +264,12 @@ func validateLocalIP(originParam string) (bool, error) {
 	return (parsedIP.IsLoopback() || parsedIP.IsPrivate()), nil
 }
 
-func validateForm(formDetails *FormDetails) bool {
+func (c *Controller) validateForm(formDetails *FormDetails) bool {
+	// Note: c.p.Validators are not checked on this level and will be checked
+	// at c.p.CreateToken
```
yolocs, line 269:
```suggestion
	// This only does sanity check of the form, e.g. field not empty.
	// The actual verification only happens when submitting the form.
```

pkg/controller/controller.go
```diff
@@ -334,6 +343,36 @@ func ttls() []string {
...
 }
 
-func categories() []string {
-	return []string{"explanation"}
+func (c *Controller) categories() []string {
+	validators := c.p.Validators()
+	categories := make([]string, 0, len(validators))
+	for v := range validators {
+		categories = append(categories, v)
+	}
```
yolocs, line 351: https://pkg.go.dev/golang.org/x/exp/maps#Keys

yolocs: Actually I'd suggest creating the map inline.

```go
for v := range validators {
  categories = append(categories, Pair{Key: v, Text: strings.ToUpper(v)}
}
```

pkg/controller/controller.go
yolocs: For now, It's only the plugin name. We can probably do https://pkg.go.dev/strings#ToTitle.

pkg/controller/controller.go
```diff
@@ -334,6 +343,36 @@ func ttls() []string {
...
+	categories = append(categories, "explanation")
```
yolocs, line 352: Such an outlier. I want to make explanation also a validator https://github.com/abcxyz/jvs/issues/302

yolocs (changes requested):

pkg/controller/controller.go
```diff
@@ -115,22 +127,22 @@ func (c *Controller) HandlePopup() http.Handler {
...
-	formDetails, err := getFormDetails(r)
+	formDetails, err := c.getFormDetails(r)
 	if err != nil {
 		c.renderBadRequest(w, err.Error())
 		return
 	}
 
 	// set some defaults for the form
-	formDetails.Category = categories()[0]
+	formDetails.Category = c.categories()[0]
```
sethvargo, line 137: I realize this is the existing code, but this can panic if the categories is empty (index out of bounds)

pkg/controller/controller.go
```diff
@@ -334,6 +343,36 @@ func ttls() []string {
 	return []string{"15m", "30m", "1h", "2h", "4h"}
 }
 
-func categories() []string {
-	return []string{"explanation"}
+func (c *Controller) categories() []string {
```
sethvargo, line 346: I'm confused about these functions:

1. Why do we need them?
2. If we do need them, can we pre-compute these values once during controller instantiation instead of doing this iteration each time? Right now, this will be called on each request, which feels rather inefficient.
3. What data/structure is actually required? We can pass maps and slices into the Go templates directly and extract keys/values in the template engine.

pkg/controller/controller.go
```diff
@@ -334,6 +343,36 @@ func ttls() []string {
...
+	}
+	categories = append(categories, "explanation")
+	return categories
+}
+
+// categoryPairs returns a list of Pairs for rendering.
+func (c *Controller) categoryPairs() []Pair {
+	pairs := make([]Pair, 0, len(c.categories()))
+	for _, category := range c.categories() {
+		pairs = append(pairs, Pair{
```
sethvargo, line 360:
```suggestion
func (c *Controller) categoryPairs() []*Pair {
	pairs := make([]Pair, 0, len(c.categories()))
	for _, category := range c.categories() {
		pairs = append(pairs, &Pair{
```

pkg/controller/controller.go
dandans-verily: Processor is from another package and requires keyManagementClient and justificationConfig to construct a new instance, so it is easier to mock it out to avoid all those complexity 

https://github.com/abcxyz/jvs/blob/c74e145411363aed1d245d5530ed92f246df1abe/pkg/justification/processor.go#L54

yolocs: Likely you can simply do

```go
p := NewProcessor(nil, &config.JustificationConfig{}).WithValidators(...)
```

pkg/controller/controller.go
dandans-verily: Cannot make this change unfortunately since []Pair is required in the place below to pass to the html

https://github.com/abcxyz/jvs/blob/c74e145411363aed1d245d5530ed92f246df1abe/pkg/controller/controller.go#L297

pkg/controller/controller.go
dandans-verily: "explanation" is hard-coded below, as a result, it won't through exception. Seems the logic is "explanation" should always be one of the category to allow user to input free style text.

https://github.com/abcxyz/jvs/blob/main/pkg/controller/controller.go#L338

pkg/controller/controller.go
dandans-verily: I have been playing with different functions under `strings` package, but don't seem a way to make things camel case.  Note that @qiaoweiz is working on passing back human-friendly category and hint from validator plugin, as a result I simply add a getText function here to stay compatible with existing code.

pkg/controller/controller.go
dandans-verily: +1 on passing back the display data from plugin.

I have been playing with different functions under `strings` package, but don't seem a way to make things camel case.  Note that @qiaoweiz is working on passing back human-friendly category and hint from validator plugin, as a result I simply add a getText function here to stay compatible with existing code.

pkg/controller/controller.go
dandans-verily: Currently categories() is used for two things:

1. set the default value of the drop down (relevant code [here](https://github.com/abcxyz/jvs/blob/main/pkg/controller/controller.go#L125))
2. validate the form and show corresponding errors (relevant code [here](https://github.com/abcxyz/jvs/blob/main/pkg/controller/controller.go#L258))

Existing code is just a hard-code string array, so it is ok to call them on every request. Good point that we should do the calculation beforehand. I have adjusted the code to calculate it when the controller is constructed.

`What data/structure is actually required? We can pass maps and slices into the Go templates directly and extract keys/values in the template engine.` <- I don't think we can pass the map and slices for the two use cases of `categories`. `getFormDetails` is doing something similar and it is the caller of `categoryPairs`

pkg/justification/processor.go
dandans-verily: Done

pkg/controller/controller.go
dandans-verily: Actually followed Seth's suggestion from https://github.com/abcxyz/jvs/pull/300#discussion_r1260028123 to precompute category and pairs in constructor, turned out mockProcessor is not needed any longer. Already cleaned up the PR, PTAL

pkg/controller/controller.go
dandans-verily: Done

pkg/controller/controller.go
dandans-verily: ack

pkg/controller/controller.go
sethvargo: We can change that here: https://github.com/abcxyz/jvs/blob/c74e145411363aed1d245d5530ed92f246df1abe/pkg/controller/controller.go#L53

pkg/controller/controller.go
sethvargo: https://github.com/abcxyz/jvs/pull/303

pkg/controller/controller.go
sethvargo: I fixed this in https://github.com/abcxyz/jvs/pull/303; we shouldn't assume the value will always be hardcoded (or have a specific length)

pkg/controller/controller.go
sethvargo: You need to use cases, not strings. I added it in #303.

yolocs: #303 is merged and hopefully it can simplify some logic in this PR

dandans-verily: Yep yep, the logic of current PR is simplified a lot now. PTAL

pkg/controller/controller.go
dandans-verily: Will use ToTitle in html for now after merging in #303

pkg/controller/controller.go
dandans-verily: Thanks a lot for fixing it. Merging #303 now.

pkg/controller/controller.go
dandans-verily: Yes, the logic is much simpler after merging in #303 . Currently I directly add more entries to the categories map, let me know what you think

pkg/controller/controller.go
dandans-verily: Resolving this comment thread now since we no longer use Pair after #303

sethvargo (approved):

yolocs (approved):

PR #307, opened 2023-07-18
feat: Update plugin interface for GetUIData

### LINKED ISSUE #305: Expose interface to get Hint and Category

qiaoweiz:
To resolve #305

### DIFF
--- a/apis/v0/plugin.go
+++ b/apis/v0/plugin.go
@@ -70,8 +70,12 @@ func (v *ExplanationValidator) Validate(_ context.Context, req *ValidateJustific
 	return &ValidateJustificationResponse{Valid: true}, nil
 }
 
+// This method retrieves UIData and handle any potential errors encountered during the process.
 func (v *ExplanationValidator) GetUIData(_ context.Context, req *GetUIDataRequest) (*UIData, error) {
-	return &UIData{}, nil
+	return &UIData{
+		DisplayName: "Explanation",
+		Hint:        "A justification reason in free-form text.",
+	}, nil
 }
 
 // ValidatorPlugin implements [plugin.GRPCPlugin].
@@ -113,6 +117,7 @@ func (m *PluginClient) Validate(ctx context.Context, req *ValidateJustificationR
 	return resp, nil
 }
 
+// This method retrieves UIData and handles any potential errors encountered during the process.
 func (m *PluginClient) GetUIData(ctx context.Context, req *GetUIDataRequest) (*UIData, error) {
 	resp, err := m.client.GetUIData(ctx, req)
 	if err != nil {
@@ -136,6 +141,7 @@ func (m *PluginServer) Validate(ctx context.Context, req *ValidateJustificationR
 	return resp, nil
 }
 
+// This method retrieves UIData and handles any potential errors encountered during the process.
 func (m *PluginServer) GetUIData(ctx context.Context, req *GetUIDataRequest) (*UIData, error) {
 	resp, err := m.Impl.GetUIData(ctx, req)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


apis/v0/plugin.go
```diff
@@ -122,3 +135,11 @@ func (m *PluginServer) Validate(ctx context.Context, req *ValidateJustificationR
 	}
 	return resp, nil
 }
+
+func (m *PluginServer) GetUIData(ctx context.Context, req *GetUIDataRequest) (*UIData, error) {
```
dandans-verily, line 139: Add documentation throughout

dandans-verily (commented):
I think you need to update pkg/controller/controller_test.go as well since it use mockValidaotor https://github.com/abcxyz/jvs/blob/9c429bb50dca9a91470d81f1f50f2140776f5e8b/pkg/justification/processor_test.go

apis/v0/plugin.go
```diff
@@ -69,6 +70,10 @@ func (v *ExplanationValidator) Validate(_ context.Context, req *ValidateJustific
 	return &ValidateJustificationResponse{Valid: true}, nil
 }
 
+func (v *ExplanationValidator) GetUIData(_ context.Context, req *GetUIDataRequest) (*UIData, error) {
+	return &UIData{}, nil
```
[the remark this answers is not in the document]
qiaoweiz: it seems be given in the [controller.go](https://github.com/abcxyz/jvs/blob/main/pkg/controller/controller.go#L136-L138). If the category is empty, then use the default value as [code](https://github.com/abcxyz/jvs/blob/main/apis/v0/plugin.go#L32). 

Please let me know if I miss anything. Thank you!

yolocs, line 74: Should return the following
Display name: "Explanation"
Hint: "A justification reason in free-form text."

yolocs (commented):
Please fix other places to make test pass

apis/v0/plugin.go
qiaoweiz: Added comments. Thank you.

apis/v0/plugin.go
qiaoweiz: Updated with default value. Thanks!

apis/v0/plugin.go
```diff
@@ -122,3 +140,12 @@ func (m *PluginServer) Validate(ctx context.Context, req *ValidateJustificationR
 	}
 	return resp, nil
 }
+
+// This method retrieves UIData and handles any potential errors encountered during the process.
```
yolocs, line 144: Go comments have convention to start with the signature name
```suggestion
// GetUIData retrieves plugin's display data.
```

qiaoweiz: Thank you! I updated the comments in the latest version.

yolocs (approved):

pkg/justification/processor_test.go
```diff
@@ -284,6 +293,27 @@ func TestCreateToken(t *testing.T) {
 			},
 			wantErr: "unexpected error from validator \"jira\": Cannot connect to validator",
 		},
+		{
+			name: "validator_missing_ui_data",
```
dandans-verily, line 297: I think we should expect error in this case instead of allowing it. If we allow it, the UI rendering will be broken and we can only detect that at run time. @yolocs what do you think?

dandans-verily: Oh, I see, this is on the processor_test level. The error I was expecting should be from controller level. And seems we are not updating the loading logic yet. Fine to add it in followed-up PR

dandans-verily (approved):

apis/v0/plugin.go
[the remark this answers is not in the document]
qiaoweiz: Updated with default value. Marked as resolved.

pkg/justification/processor_test.go
qiaoweiz: Marked as resolved. Thank you!

PR #309, opened 2023-07-19
feat: support different justification category in CLI

yolocs:
Fix: https://github.com/abcxyz/jvs/issues/278

But should we consider a command to list available category? That will require a new JVS API...

### DIFF
--- a/pkg/cli/token_create.go
+++ b/pkg/cli/token_create.go
@@ -134,7 +134,8 @@ func (c *TokenCreateCommand) Flags() *cli.FlagSet {
 		Target:  &c.flagExplanation,
 		Aliases: []string{"e"},
 		Example: "Debugging ticket #123",
-		Usage:   `DEPRECATED. Use "justification" flag instead. This flag will be removed in a future release.`,
+		Usage:   `A reason for the action.`,
+		Hidden:  true,
 	})
 
 	f.StringVar(&cli.StringVar{
@@ -147,7 +148,7 @@ func (c *TokenCreateCommand) Flags() *cli.FlagSet {
 
 	f.StringVar(&cli.StringVar{
 		Name:    "justification",
-		Target:  &c.flagExplanation,
+		Target:  &c.flagJustificationText,
 		Aliases: []string{"j"},
 		Usage:   `The justification text. The format depends on the justification category.`,
 	})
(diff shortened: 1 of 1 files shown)


pkg/cli/token_create.go
```diff
@@ -125,7 +134,22 @@ func (c *TokenCreateCommand) Flags() *cli.FlagSet {
...
+		Name:    "category",
+		Target:  &c.flagCategory,
+		EnvVar:  "JVSCTL_CATEGORY",
+		Default: jvspb.DefaultJustificationCategory,
+		Usage:   `The justification category.`,
+	})
+
+	f.StringVar(&cli.StringVar{
+		Name:    "justification",
+		Target:  &c.flagExplanation,
```
sqin2019, line 150: flagJustificationText?

pkg/cli/token_create.go
```diff
@@ -185,8 +209,16 @@ func (c *TokenCreateCommand) Run(ctx context.Context, args []string) error {
 		return fmt.Errorf("unexpected arguments: %q", args)
 	}
 
-	if c.flagExplanation == "" {
-		return fmt.Errorf("explanation is required")
+	if c.flagExplanation != "" {
+		fmt.Fprintln(c.Stderr(), `WARNING: "explanation" flag is DEPRECATED and will be removed in a future release. Use "justification" flag instead.`)
```
sethvargo, line 213:
```suggestion
		c.Errf(`WARNING: the "-explanation" flag is deprecated and will be removed in a future release. Use "-justification" instead.`)
```

pkg/cli/token_create.go
sethvargo, line 150:
```suggestion
		Target:  &c.flagJustificationText,
```

sethvargo (changes requested):

pkg/cli/token_create.go
```diff
@@ -125,7 +134,22 @@ func (c *TokenCreateCommand) Flags() *cli.FlagSet {
 		Target:  &c.flagExplanation,
 		Aliases: []string{"e"},
 		Example: "Debugging ticket #123",
-		Usage:   `A reason for the action.`,
+		Usage:   `DEPRECATED. Use "justification" flag instead. This flag will be removed in a future release.`,
```
sethvargo, line 137: I'd probably leave the message as-is and hide the flag: https://github.com/abcxyz/pkg/blob/05a9df35d989594e4f242e397f29d96f3fc0b041/cli/flags.go#L542

```suggestion
Hidden: true,
Usage: `A reason for the action.`,
```

sethvargo (approved):

Issue #310, opened 2023-07-19
Render additional Hint UI


qiaoweiz:
### TL;DR

This task includes two parts:

- Adjust popup.html and main.js logic so hint will change when selected option is changed
- Adjust controller.go to pass in Hint and category sourced from plugin

PR #311, opened 2023-07-19
feat: Adjust controller.go to pass in UIData sourced from plugin

qiaoweiz:
To resolve: #310

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -92,8 +92,8 @@ type ErrorDetails struct {
 
 const iapHeaderName = "x-goog-authenticated-user-email"
 
-func New(h *renderer.Renderer, p *justification.Processor, allowlist []string, ctx context.Context) (*Controller, error) {
-	categories, err := getCatagoriesDisplayData(p.Validators(), ctx)
+func New(ctx context.Context, h *renderer.Renderer, p *justification.Processor, allowlist []string) (*Controller, error) {
+	categories, err := catagoriesDisplayData(ctx, p.Validators())
 	if err != nil {
 		return nil, err
 	}
@@ -343,14 +343,14 @@ func getEmail(r *http.Request) (string, error) {
 	return split[1], nil
 }
 
-// GetCatagoriesDisplayData takes validators as input and returns UIData for display. If there's an error during the GetUIData RPC call, it will be thrown.
-func getCatagoriesDisplayData(validators map[string]jvspb.Validator, ctx context.Context) (map[string]*jvspb.UIData, error) {
+// categoriesDisplayData gathers the plugins' display data.
+func catagoriesDisplayData(ctx context.Context, validators map[string]jvspb.Validator) (map[string]*jvspb.UIData, error) {
 	displayData := make(map[string]*jvspb.UIData, len(validators))
 
 	for k, v := range validators {
 		d, err := v.GetUIData(ctx, &jvspb.GetUIDataRequest{})
 		if err != nil {
-			return nil, fmt.Errorf("failed to get UIData with error %w", err)
+			return nil, fmt.Errorf("failed to get display data for category %q: %w", k, err)
 		}
 		displayData[k] = d
 	}
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -92,8 +92,8 @@ type ErrorDetails struct {
 
 const iapHeaderName = "x-goog-authenticated-user-email"
 
-func New(h *renderer.Renderer, p *justification.Processor, allowlist []string) (*Controller, error) {
-	categories, err := getCatagoriesDisplayData(p.Validators())
+func New(h *renderer.Renderer, p *justification.Processor, allowlist []string, ctx context.Context) (*Controller, error) {
+	categories, err := getCatagoriesDisplayData(p.Validators(), ctx)
 	if err != nil {
 		return nil, err
 	}
@@ -344,11 +344,11 @@ func getEmail(r *http.Request) (string, error) {
 }
 
 // GetCatagoriesDisplayData takes validators as input and returns UIData for display. If there's an error during the GetUIData RPC call, it will be thrown.
-func getCatagoriesDisplayData(validators map[string]jvspb.Validator) (map[string]*jvspb.UIData, error) {
-	displayData := make(map[string]*jvspb.UIData)
+func getCatagoriesDisplayData(validators map[string]jvspb.Validator, ctx context.Context) (map[string]*jvspb.UIData, error) {
+	displayData := make(map[string]*jvspb.UIData, len(validators))
 
 	for k, v := range validators {
-		d, err := v.GetUIData(context.Background(), &jvspb.GetUIDataRequest{})
+		d, err := v.GetUIData(ctx, &jvspb.GetUIDataRequest{})
 		if err != nil {
 			return nil, fmt.Errorf("failed to get UIData with error %w", err)
 		}
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -47,7 +47,7 @@ type Controller struct {
 	h                   *renderer.Renderer
 	p                   *justification.Processor
 	allowlist           []string
-	categoryDisplayData map[string]struct{}
+	categoryDisplayData map[string]*jvspb.UIData
 }
 
 // Content defines the displayable parts of the token retrieval form.
@@ -56,7 +56,7 @@ type Content struct {
 	CategoryLabel string
 	ReasonLabel   string
 	TTLLabel      string
-	Categories    map[string]struct{}
+	Categories    map[string]*jvspb.UIData
 	TTLs          map[string]struct{}
 }
 
@@ -92,17 +92,18 @@ type ErrorDetails struct {
 
 const iapHeaderName = "x-goog-authenticated-user-email"
 
-func New(h *renderer.Renderer, p *justification.Processor, allowlist []string) *Controller {
-	categories := make(map[string]struct{})
-	for v := range p.Validators() {
-		categories[v] = struct{}{}
+func New(h *renderer.Renderer, p *justification.Processor, allowlist []string) (*Controller, error) {
+	categories, err := getCatagoriesDisplayData(p.Validators())
+	if err != nil {
+		return nil, err
 	}
+
 	return &Controller{
 		h:                   h,
 		p:                   p,
 		allowlist:           allowlist,
 		categoryDisplayData: categories,
-	}
+	}, nil
 }
 
 func (c *Controller) HandleHealth() http.Handler {
@@ -340,4 +341,18 @@ func getEmail(r *http.Request) (string, error) {
 	}
 
 	return split[1], nil
+}
+
+// GetCatagoriesDisplayData takes validators as input and returns UIData for display. If there's an error during the GetUIData RPC call, it will be thrown.
+func getCatagoriesDisplayData(validators map[string]jvspb.Validator) (map[string]*jvspb.UIData, error) {
+	displayData := make(map[string]*jvspb.UIData)
+
+	for k, v := range validators {
+		d, err := v.GetUIData(context.Background(), &jvspb.GetUIDataRequest{})
+		if err != nil {
+			return nil, fmt.Errorf("failed to get UIData with error %w", err)
+		}
+		displayData[k] = d
+	}
+	return displayData, nil
 }
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -104,7 +104,7 @@ func TestHandlePopup(t *testing.T) {
 			t.Parallel()
 
 			harness := envtest.NewServerConfig(t, "9091", tc.allowlist, true)
-			c, _ := New(harness.Renderer, harness.Processor, tc.allowlist)
+			c, _ := New(harness.Renderer, harness.Processor, tc.allowlist, context.Background())
 
 			w, r := envtest.BuildFormRequest(ctx, t, tc.method, tc.path,
 				tc.queryParam)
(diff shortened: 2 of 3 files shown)


pkg/controller/controller.go
```diff
@@ -341,3 +342,17 @@ func getEmail(r *http.Request) (string, error) {
 
 	return split[1], nil
 }
+
+// GetCatagoriesDisplayData takes validators as input and returns UIData for display. If there's an error during the GetUIData RPC call, it will be thrown.
+func getCatagoriesDisplayData(validators map[string]jvspb.Validator) (map[string]*jvspb.UIData, error) {
+	displayData := make(map[string]*jvspb.UIData)
+
+	for k, v := range validators {
+		d, err := v.GetUIData(context.Background(), &jvspb.GetUIDataRequest{})
```
dandans-verily, line 351: context should be passed in from [here](https://github.com/abcxyz/jvs/blob/3e1a165416e738a0b8ab6da57577b21ef8d36b2e/pkg/ui/server.go#L52). Actually I am not sure whether rpc call can be made at this stage, should it wait till the server and plugin are started successfully? Another option is to populate it at the first call. @yolocs what do you think?

pkg/controller/controller.go
```diff
@@ -92,17 +92,18 @@ type ErrorDetails struct {
 
 const iapHeaderName = "x-goog-authenticated-user-email"
 
-func New(h *renderer.Renderer, p *justification.Processor, allowlist []string) *Controller {
-	categories := make(map[string]struct{})
-	for v := range p.Validators() {
-		categories[v] = struct{}{}
+func New(h *renderer.Renderer, p *justification.Processor, allowlist []string) (*Controller, error) {
+	categories, err := getCatagoriesDisplayData(p.Validators())
```
sethvargo, line 96: 👍 good call on pre-computing this at initialization.

pkg/controller/controller.go
```diff
@@ -341,3 +342,17 @@ func getEmail(r *http.Request) (string, error) {
 
 	return split[1], nil
 }
+
+// GetCatagoriesDisplayData takes validators as input and returns UIData for display. If there's an error during the GetUIData RPC call, it will be thrown.
+func getCatagoriesDisplayData(validators map[string]jvspb.Validator) (map[string]*jvspb.UIData, error) {
+	displayData := make(map[string]*jvspb.UIData)
```
sethvargo, line 348: Nit: you know how big this will be, save the allocation:

```suggestion
	displayData := make(map[string]*jvspb.UIData, len(validators))
```

pkg/controller/controller.go
sethvargo: Your initial point is correct - this function should take the `ctx` from `New` and thread it through here.

sethvargo, line 351: What is the second argument here?

qiaoweiz: Could you kindly clarify if you were inquiring about "&jvspb.GetUIDataRequest{}"? It was introduced in this [pr](https://github.com/abcxyz/jvs/commit/9c429bb50dca9a91470d81f1f50f2140776f5e8b).

sethvargo: Sorry, I meant that it's a little strange that it's an empty struct. It doesn't look like there's any fields for GetUIDataRequest in the proto though, so maybe that's WAI.

qiaoweiz: SG.

pkg/controller/controller.go
qiaoweiz: Updated in the latest version.

pkg/controller/controller.go
qiaoweiz: Thank you for your suggestion! 

Updated in the latest version to take ctx from New method from [here](https://github.com/abcxyz/jvs/blob/3e1a165416e738a0b8ab6da57577b21ef8d36b2e/pkg/ui/server.go#L52)

pkg/controller/controller.go
```diff
@@ -92,17 +92,18 @@ type ErrorDetails struct {
 
 const iapHeaderName = "x-goog-authenticated-user-email"
 
-func New(h *renderer.Renderer, p *justification.Processor, allowlist []string) *Controller {
-	categories := make(map[string]struct{})
-	for v := range p.Validators() {
-		categories[v] = struct{}{}
+func New(h *renderer.Renderer, p *justification.Processor, allowlist []string, ctx context.Context) (*Controller, error) {
```
yolocs, line 95: Context should pretty much always be the first parameter

pkg/controller/controller.go
```diff
@@ -92,8 +92,8 @@ type ErrorDetails struct {
 
 const iapHeaderName = "x-goog-authenticated-user-email"
 
-func New(h *renderer.Renderer, p *justification.Processor, allowlist []string) (*Controller, error) {
-	categories, err := getCatagoriesDisplayData(p.Validators())
+func New(h *renderer.Renderer, p *justification.Processor, allowlist []string, ctx context.Context) (*Controller, error) {
```
sethvargo, line 95: Context should always be the _first_ parameter to a function.

```suggestion
func New(ctx context.Context, h *renderer.Renderer, p *justification.Processor, allowlist []string) (*Controller, error) {
```

pkg/controller/controller.go
```diff
@@ -344,11 +344,11 @@ func getEmail(r *http.Request) (string, error) {
 }
 
 // GetCatagoriesDisplayData takes validators as input and returns UIData for display. If there's an error during the GetUIData RPC call, it will be thrown.
-func getCatagoriesDisplayData(validators map[string]jvspb.Validator) (map[string]*jvspb.UIData, error) {
-	displayData := make(map[string]*jvspb.UIData)
+func getCatagoriesDisplayData(validators map[string]jvspb.Validator, ctx context.Context) (map[string]*jvspb.UIData, error) {
```
sethvargo, line 347: Context should always be the _first_ parameter to a function.


```suggestion
func getCatagoriesDisplayData(ctx context.Context, validators map[string]jvspb.Validator) (map[string]*jvspb.UIData, error) {
```

sethvargo (changes requested):

pkg/controller/controller.go
```diff
@@ -341,3 +342,17 @@ func getEmail(r *http.Request) (string, error) {
 
 	return split[1], nil
 }
+
+// GetCatagoriesDisplayData takes validators as input and returns UIData for display. If there's an error during the GetUIData RPC call, it will be thrown.
+func getCatagoriesDisplayData(validators map[string]jvspb.Validator, ctx context.Context) (map[string]*jvspb.UIData, error) {
```
yolocs, line 347: Make context the first parameter.

In Go, we usually don't name func as `getXxx`. `categoriesDisplayData`

For private func, we don't have to follow Go comment convention. But if you want, you can still do `// categoriesDisplayData gathers the plugins' display data`. Also, Go always explicitly returns error so no need to call that out.

pkg/controller/controller.go
```diff
@@ -341,3 +342,17 @@ func getEmail(r *http.Request) (string, error) {
...
+	displayData := make(map[string]*jvspb.UIData, len(validators))
+
+	for k, v := range validators {
+		d, err := v.GetUIData(ctx, &jvspb.GetUIDataRequest{})
+		if err != nil {
+			return nil, fmt.Errorf("failed to get UIData with error %w", err)
```
yolocs, line 353:
```suggestion
			return nil, fmt.Errorf("failed to get display data for category %q: %w", k, err)
```

pkg/controller/controller_test.go
```diff
@@ -104,7 +104,7 @@ func TestHandlePopup(t *testing.T) {
 			t.Parallel()
 
 			harness := envtest.NewServerConfig(t, "9091", tc.allowlist, true)
-			c := New(harness.Renderer, harness.Processor, tc.allowlist)
+			c, _ := New(harness.Renderer, harness.Processor, tc.allowlist, context.Background())
```
yolocs, line 107: Never ignore error.
```suggestion
			c, err := New(harness.Renderer, harness.Processor, tc.allowlist, context.Background())
			if err != nil {
			  t.Fatal(err)
			}
```

pkg/controller/controller_test.go
```diff
@@ -291,7 +291,7 @@ func TestValidateForm(t *testing.T) {
 		},
 	})
 
-	controller := New(nil, p, []string{})
+	controller, _ := New(nil, p, []string{}, context.Background())
```
yolocs, line 294: Don't ignore error

pkg/controller/controller_test.go
```diff
@@ -457,3 +457,81 @@ func TestGetEmail(t *testing.T) {
...
+		Hint:        "Jira Issue key under JVS project",
+	}
+	gitUIData := &jvspb.UIData{
+		DisplayName: "Git issue key",
+		Hint:        "Git Issue key under JVS project",
+	}
+	expectedCatagoriesData := map[string]*jvspb.UIData{
+		"jira": jiraUIData,
+		"git":  gitUIData,
+	}
```
yolocs, line 487: Inline all these in line 503.

pkg/controller/controller_test.go
```diff
@@ -457,3 +457,81 @@ func TestGetEmail(t *testing.T) {
...
+
+			for k, w := range tc.wantRes {
+				g, ok := gotRes[k]
+				if !ok {
+					t.Errorf("Key '%s' not found in the actual map", k)
+				}
+
+				if w.DisplayName != g.DisplayName || w.Hint != g.Hint {
+					t.Errorf("Values for key '%s' are different. Expected: %v, Actual: %v", k, w, g)
+				}
```
yolocs, line 533: Use `cmp`.

```go
if diff := cmp.Diff(want, got, protocmp.Transform()); diff != "" {
  t.Errorf("Display data (-want,+got):\n%s", diff)
}
```

pkg/ui/server.go
```diff
@@ -48,8 +48,13 @@ func NewServer(ctx context.Context, uiCfg *config.UIServiceConfig, p *justificat
 		return nil, fmt.Errorf("failed to create renderer: %w", err)
 	}
 
+	uic, cerr := controller.New(h, p, uiCfg.Allowlist, ctx)
+	if cerr != nil {
+		return nil, fmt.Errorf("failed to create controller: %w", cerr)
+	}
```
yolocs, line 54: No need to use a different err name here.
```suggestion
	uic, err := controller.New(h, p, uiCfg.Allowlist, ctx)
	if err != nil {
		return nil, fmt.Errorf("failed to create controller: %w", err)
	}
```

yolocs (changes requested):

pkg/controller/controller.go
qiaoweiz: Thank you.

pkg/controller/controller.go
qiaoweiz: Updated in the latest version. Thank you.

pkg/controller/controller.go
qiaoweiz: Updated in the latest version. Thank you.

pkg/controller/controller.go
qiaoweiz: Updated in the latest version. Thank you.

pkg/controller/controller.go
qiaoweiz: Updated the func name to categoriesDisplayData
and updated the comment as your suggestion.

pkg/controller/controller.go
qiaoweiz: Updated in the latest version.

pkg/controller/controller_test.go
qiaoweiz: Updated in the latest version.

pkg/controller/controller_test.go
qiaoweiz: Updated in the latest version as you suggested in another comment.

pkg/controller/controller_test.go
qiaoweiz: Moved all these into the second test case

pkg/controller/controller_test.go
qiaoweiz: Thank you for the suggestion. I updated this part in the latest version.

pkg/ui/server.go
qiaoweiz: Updated in the latest version.

pkg/controller/controller.go
```diff
@@ -92,17 +92,18 @@ type ErrorDetails struct {
 
 const iapHeaderName = "x-goog-authenticated-user-email"
 
-func New(h *renderer.Renderer, p *justification.Processor, allowlist []string) *Controller {
-	categories := make(map[string]struct{})
-	for v := range p.Validators() {
-		categories[v] = struct{}{}
+func New(ctx context.Context, h *renderer.Renderer, p *justification.Processor, allowlist []string) (*Controller, error) {
+	categories, err := catagoriesDisplayData(ctx, p.Validators())
```
dandans-verily, line 96: Sanity check whether the plugin is ready to serve data at initialization. Could you run it locally to ensure it is working?

pkg/controller/controller_test.go
```diff
@@ -104,7 +105,10 @@ func TestHandlePopup(t *testing.T) {
 			t.Parallel()
 
 			harness := envtest.NewServerConfig(t, "9091", tc.allowlist, true)
-			c := New(harness.Renderer, harness.Processor, tc.allowlist)
+			c, err := New(context.Background(), harness.Renderer, harness.Processor, tc.allowlist)
```
sethvargo, line 108: ctx is defined on L60/61. You can use that here.

pkg/controller/controller.go
dandans-verily: Confirmed offline that it would work :D Resolving this thread now

pkg/controller/controller_test.go
qiaoweiz: updated in the latest version.

yolocs (approved):

dandans-verily (approved):

sethvargo (approved):

PR #313, opened 2023-07-21
Show Hint from UiData On Popup UI

### LINKED ISSUE #312: Render additional Hint UI
### TL;DR

Render additional Hint UI

### Detailed design

_No response_

### Alternatives considered

_No response_

### Additional information

_No response_

qiaoweiz:
resolve #312

### DIFF
--- a/assets/server/popup.html
+++ b/assets/server/popup.html
@@ -25,11 +25,7 @@
           <select class="content-select" id="category" name="category">
             {{ range $element, $value := .Content.Categories }}
             {{ if eq $element $context.Category }}
-            {{ $hint := $value.Hint }}
-            {{ if or (not $hint) (eq $hint "") }}
-              {{ $hint = "i.e. issue/xxxxx" }}
-            {{ end }}
-            <option value="{{ $element }}" selected="selected" hint="{{ $hint }}">{{ $element | toTitle }}</option>
+            <option value="{{ $element }}" selected="selected" hint="{{ $value.Hint }}">{{ $element | toTitle }}</option>
             {{ else }}
             <option value="{{ $element }}">{{ $element }}</option>
             {{ end }}
--- a/assets/server/popup.html
+++ b/assets/server/popup.html
@@ -25,7 +25,11 @@
           <select class="content-select" id="category" name="category">
             {{ range $element, $value := .Content.Categories }}
             {{ if eq $element $context.Category }}
-            <option value="{{ $element }}" selected="selected">{{ $element | toTitle }}</option>
+            {{ $hint := $value.Hint }}
+            {{ if or (not $hint) (eq $hint "") }}
+              {{ $hint = "i.e. issue/xxxxx" }}
+            {{ end }}
+            <option value="{{ $element }}" selected="selected" hint="{{ $hint }}">{{ $element | toTitle }}</option>
             {{ else }}
             <option value="{{ $element }}">{{ $element }}</option>
             {{ end }}
--- a/assets/server/static/js/popup/main.js
+++ b/assets/server/static/js/popup/main.js
@@ -12,13 +12,23 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-document.addEventListener("DOMContentLoaded", function () {
-  var categorySelect = document.getElementById("category");
-  var reasonInput = document.getElementById("reason");
+document.addEventListener("DOMContentLoaded", async function () {
+  const categorySelect = document.querySelector("#category");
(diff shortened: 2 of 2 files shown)


assets/server/static/js/popup/main.js
```diff
@@ -12,6 +12,23 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+document.addEventListener("DOMContentLoaded", function () {
```
sethvargo, line 15:
```suggestion
document.addEventListener("DOMContentLoaded", async function () {
```

assets/server/static/js/popup/main.js
```diff
@@ -12,6 +12,23 @@
...
+  var categorySelect = document.getElementById("category");
+  var reasonInput = document.getElementById("reason");
```
sethvargo, line 17: Prefer `document.querySelector`, use `const/let` instead of `var`:

```suggestion
  const categorySelect = document.querySelector("#category");
  const reasonInput = document.querySelector("#reason");
```

assets/server/static/js/popup/main.js
```diff
@@ -12,6 +12,23 @@
...
+
+  // Update the reason input's placeholder with the selected category's hint.
+  function updatePlaceholder() {
+    var selectedOption = categorySelect.options[categorySelect.selectedIndex];
+    reasonInput.placeholder = selectedOption.getAttribute("hint");
+  }
+
+  // Call the function when the page loads.
+  updatePlaceholder();
```
sethvargo, line 26: You should guard against these objects not being found before calling `updatePlaceholder`.

assets/server/static/js/popup/main.js
```diff
@@ -12,6 +12,23 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+document.addEventListener("DOMContentLoaded", function () {
+  var categorySelect = document.getElementById("category");
+  var reasonInput = document.getElementById("reason");
+
+  // Update the reason input's placeholder with the selected category's hint.
+  function updatePlaceholder() {
+    var selectedOption = categorySelect.options[categorySelect.selectedIndex];
```
sethvargo, line 21: Prefer `const/let`:

```suggestion
    const selectedOption = categorySelect.options[categorySelect.selectedIndex];
```

assets/server/popup.html
```diff
@@ -25,7 +25,11 @@ <h1 class="title">{{ .PageTitle }}</h1>
           <select class="content-select" id="category" name="category">
             {{ range $element, $value := .Content.Categories }}
             {{ if eq $element $context.Category }}
-            <option value="{{ $element }}" selected="selected">{{ $element | toTitle }}</option>
+            {{ $hint := $value.Hint }}
+            {{ if or (not $hint) (eq $hint "") }}
+              {{ $hint = "i.e. issue/xxxxx" }}
+            {{ end }}
+            <option value="{{ $element }}" selected="selected" hint="{{ $hint }}">{{ $element | toTitle }}</option>
```
sethvargo, line 32: `not $hint` should cover the empty string case too, thus this can be simplified:

```suggestion
            <option value="{{ $element }}" selected="selected" hint="{{ or $value.Hint "i.e. issue/xxxx" }}">{{ $element | toTitle }}</option>
```

However, I feel like we should _always_ set a default hint in the controller to avoid this logic in the template.

sethvargo (changes requested):

assets/server/popup.html
yolocs: I agree. All defaults should come from https://github.com/abcxyz/jvs/blob/main/apis/v0/plugin.go#L74-L79

yolocs: To clarify, I think we should just remove this special hint logic. When a category is selected, its corresponding hint should show up.

assets/server/static/js/popup/main.js
qiaoweiz: updated in the latest version.

assets/server/static/js/popup/main.js
qiaoweiz: updated in the latest version.

assets/server/static/js/popup/main.js
qiaoweiz: updated in the latest version.

assets/server/popup.html
qiaoweiz: removed this logic in the latest version.

sethvargo (approved):

PR #314, opened 2023-07-21
add example to release plugin on top of jvs image

dandans-verily:
Link to: #282


yolocs (changes requested):
I don't think this should live in the jvs repo. The plguin repo might be a more proper place. Let's discuss.

dandans-verily:
> 

I was considering that option as well, but if I put it there, it would be jiraPlugin+jvsHost. For the cases where multiple plugins are needed (e.g. both jiraPlugin and gitHubIssuePlugin), something still need to be provided in this repo

yolocs:
> > 
> 
> I was considering that option as well, but if I put it in jira repo, it can only support one plugin. For the cases where multiple plugins are needed (e.g. both jiraPlugin and gitHubIssuePlugin), something still need to be provided in this repo

For multi-plugin case, let's defer the decision of where to put it. JVS should provide the base (binary or image) instead of releasing every plugin combination.

dandans-verily:
> > > 
> > 
> > 
> > I was considering that option as well, but if I put it in jira repo, it can only support one plugin. For the cases where multiple plugins are needed (e.g. both jiraPlugin and gitHubIssuePlugin), something still need to be provided in this repo
> 
> For multi-plugin case, let's defer the decision of where to put it. JVS should provide the base (binary or image) instead of releasing every plugin combination.

ok, I am closing this PR now and will send out the one from jira repo

PR #317, opened 2023-07-26
add terraform to take in jira plugin variables

dandans-verily:
Close: #280 , #279

### DIFF
--- a/terraform/modules/jvs-services/variables.tf
+++ b/terraform/modules/jvs-services/variables.tf
@@ -109,3 +109,16 @@ variable "ui_envvars" {
     "JVS_UI_ALLOWLIST" = "*",
   }
 }
+
+variable "jira_plugin_envvars" {
+  description = "Env vars for Jira plugin."
+  type        = map(string)
+  default = {
+    "JIRA_PLUGIN_ENDPOINT" : "https://blahblah.atlassian.net/rest/api/3",
+    "JIRA_PLUGIN_JQL" : "project = JRA and assignee != jsmith",
+    "JIRA_PLUGIN_ACCOUNT" : "abc@xyz.com",
+    "JIRA_PLUGIN_API_TOKEN_SECRET_ID" : "projects/123456/secrets/api-token/versions/4",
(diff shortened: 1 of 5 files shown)


terraform/modules/jvs-services/variables.tf
```diff
@@ -109,3 +109,16 @@ variable "ui_envvars" {
...
+variable "jira_plugin_envvars" {
+  description = "Env vars for Jira plugin."
+  type        = map(string)
+  default = {
+    "JIRA_PLUGIN_ENDPOINT" : "https://blahblah.atlassian.net/rest/api/3",
+    "JIRA_PLUGIN_JQL" : "project = JRA and assignee != jsmith",
+    "JIRA_PLUGIN_ACCOUNT" : "abc@xyz.com",
+    "JIRA_PLUGIN_API_TOKEN_SECRET_ID" : "projects/123456/secrets/api-token/versions/4",
+    "JIRA_PLUGIN_DISPLAY_NAME" : "Jira Issue Key",
+    "JIRA_PLUGIN_HINT" : "Jira Issue Key under JVS project",
```
sqin2019, line 122: Are these really meaningful? since this variable jira_plugin_envvars is optional (with default), is my understanding correct: if jira_plugin_jira is enabled, use this var to pass in envvars, if not, ignore and default with be empty?

yolocs (commented):
I'm not sure I understand where this PR leads to. The main JVS repo shouldn't depend on any plugin directly.

terraform/modules/jvs-services/variables.tf
dandans-verily: Those are just example value actually. Chatted offline with @yolocs , will keep it generic here and provide example in `jvs-plugin-jira` instead

dandans-verily:
> 

Chatted offline, will define `plugin_envvar` here and add installation guide in jvs-plugin-jira repo instead.

sqin2019 (approved):

yolocs (approved):

PR #320, opened 2023-08-01
output api and ui service accounts

dandans-verily:
- Output api and ui service accounts so they can be used by https://github.com/abcxyz/jvs-plugin-jira/pull/31
- Clean up prober_audience
- Link to: https://github.com/abcxyz/jvs-plugin-jira/issues/28

### DIFF
--- a/terraform/e2e/main.tf
+++ b/terraform/e2e/main.tf
@@ -89,5 +89,6 @@ module "jvs_monitoring" {
   prober_jvs_api_address         = "${var.jvs_api_domain}:443"
   prober_jvs_public_key_endpoint = "https://${var.jvs_api_domain}/.well-known/jwks"
   jvs_prober_image               = var.jvs_prober_image
-  prober_audience                = module.jvs_services.jvs_api_service_url
+  # Default to JVS API backend url.
+  prober_audience = var.prober_audience == "" ? module.jvs_services.jvs_api_service_url : var.prober_audience
 }
--- a/terraform/e2e/main.tf
+++ b/terraform/e2e/main.tf
@@ -89,5 +89,5 @@ module "jvs_monitoring" {
   prober_jvs_api_address         = "${var.jvs_api_domain}:443"
   prober_jvs_public_key_endpoint = "https://${var.jvs_api_domain}/.well-known/jwks"
   jvs_prober_image               = var.jvs_prober_image
-  prober_audience                = var.prober_audience
+  prober_audience                = module.jvs_services.jvs_api_service_url
 }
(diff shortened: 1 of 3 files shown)


terraform/e2e/main.tf
```diff
@@ -89,5 +89,5 @@ module "jvs_monitoring" {
   prober_jvs_api_address         = "${var.jvs_api_domain}:443"
   prober_jvs_public_key_endpoint = "https://${var.jvs_api_domain}/.well-known/jwks"
   jvs_prober_image               = var.jvs_prober_image
-  prober_audience                = var.prober_audience
+  prober_audience                = module.jvs_services.jvs_api_service_url
```
sqin2019, line 92: Thanks for correcting this! @sailorlqh Since you verified in the past, could you confirm if there is any special format for this value? Also note here - we will need to update our dev project after this.

sethvargo (dismissed):

terraform/e2e/main.tf
yolocs: Can we keep the input but make it optional (also update the input doc)?

```
# Default to JVS API backend url.
prober_audience = var.prober_audience == "" ? module.jvs_services.jvs_api_service_url : var.prober_audience
```

We really should use the custom domain instead. 

We shouldn't assume the backend service url.
But we will need to update cloud run module to allow setting custom audience https://cloud.google.com/run/docs/configuring/custom-audiences

yolocs (changes requested):

terraform/e2e/main.tf
yolocs: And we shouldn't change how we probe in dev project. gcloud app id is what a real user would get when they generate idtoken from gcloud.

sqin2019: Could you explain a little bit more on why the idtoken is generated from gcloud?

yolocs: See what audience you have from `gcloud auth print-identity-token`

sqin2019: I see, I was wondering about the [prober id_token](https://github.com/abcxyz/jvs/blob/main/prober/prober.sh#L5). So ideally jvs api will have custom domain audience for users using jvsctl, and prober is reusing this audience, is that what you mean?

yolocs: You (a human) cannot not set the audience in the idtoken if you're not impersonating a service account. Meaning, as a human using CLI, you can only (easily) get an idtoken with gcloud app id.

Our prober (who is a service account) is using the same audience as a human would do even though it can use a different audience.

sqin2019: Got it, thanks! since the custom audience is still Pre-GA, either way is fine to me.

dandans-verily: Thanks both for the great info. Updated the PR to make `prober_audience` an optional variable.

sethvargo (approved):

yolocs (approved):

PR #330, opened 2023-08-17
Update to use slog

sethvargo:


pkg/cli/public_key_server.go
```diff
@@ -62,10 +62,12 @@ func (c *PublicKeyServerCommand) Flags() *cli.FlagSet {
 }
 
 func (c *PublicKeyServerCommand) Run(ctx context.Context, args []string) error {
+	logger := logging.FromContext(ctx)
+
 	server, mux, closer, err := c.RunUnstarted(ctx, args)
 	defer func() {
 		if err := closer.Close(); err != nil {
-			logging.FromContext(ctx).Errorw("failed to close", "error", err)
```
drevell, line 68: Out of curiosity, do we actually use any of the ctx information? Would `logger.Error()` suffice? It does seem a bit verbose. I'm new to slog so this might not be a good question.

drevell (approved):

pkg/cli/public_key_server.go
sethvargo: Per https://github.com/abcxyz/abc-demo/blob/main/main.go:

> It is recommended to pass a context to an output method if one is available.


In the future, if we ever adopted something like opencensus, we could update our handler to extract data from the provided context (such as span data), and inject it into the log. You are correct that it's not currently in use.

drevell (approved):

Issue #318, opened 2023-07-27
Append Jira Issue Id and Link to Justification


qiaoweiz:
**Background:** 

Identification for Jira Issue can be achieved using either the Jira Issue Id or Jira issue key. However, it is important to note that the Jira issue key is not stable and can change if the issue is moved to a different project. On the other hand, the Jira Issue Id remains stable, but acquiring it through the UI interface is challenging. As a solution, we plan to modify the Jira plugin interface to include the Jira Issue Id in the Justification. In addition to that, a clickable link will prove to be valuable as it allows auditors to jump directly to the relevant issue, streamlining the audit process and enhancing efficiency.

**Implementation Plan:**

- Step1: Adjust Plugin interface to allow return Jira Issue Id
  - Update ValidateJustificationResponse proto with a new field, annotation. This is to store additional info the plugin might want to encapsulate in the Justification. 
  - Update [abcxyz/jvs-plugin-jira](https://github.com/abcxyz/jvs-plugin-jira).
 - Step2: Append Jira Issue Id and Issue Clickable Link to Justification
 - Step3: Testing and update Lumberjack
 - Step4: Deploy to Plato

yolocs:
Could you copy over the high level change we're proposing and why it will be useful? This will help preserve the context in this repo.

sqin2019:
@qiaoweiz closing, feel free to reopen.

PR #331, opened 2023-08-18
feat: Update justification proto with annotation field

qiaoweiz:
part of #318


sethvargo (commented):
This seems fine - do we need to do anything to update the java protos @yolocs?

yolocs:
> This seems fine - do we need to do anything to update the java protos @yolocs?

Java client doesn't use proto. It just validates the token.

yolocs (approved):

PR #335, opened 2023-08-29
feat: copy the annotation from ValidateJustificationResponse to Justification

qiaoweiz:
Tested in local. The annotation can be in the Justification token now.

### DIFF
--- a/pkg/justification/processor_test.go
+++ b/pkg/justification/processor_test.go
@@ -60,6 +60,20 @@ func (m *mockValidator) GetUIData(ctx context.Context, req *jvspb.GetUIDataReque
 	return m.uiData, m.err
 }
 
+var explanationJustification = []*jvspb.Justification{
+	{
+		Category: "explanation",
+		Value:    "test",
+	},
+}
+
+var jiraJustification = []*jvspb.Justification{
+	{
+		Category: "jira",
+		Value:    "test",
+	},
+}
+
 func TestCreateToken(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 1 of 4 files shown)


pkg/justification/processor_test.go
```diff
@@ -222,6 +222,10 @@ func TestCreateToken(t *testing.T) {
 				"jira": &mockValidator{
 					resp: &jvspb.ValidateJustificationResponse{
 						Valid: true,
+						Annotation: map[string]string{
```
sqin2019, line 225: Missing verifying wantAnnotations.

pkg/justification/processor_test.go
```diff
@@ -166,6 +191,12 @@ func TestCreateToken(t *testing.T) {
 			},
 			wantTTL:       15 * time.Minute, // comes from default
 			wantAudiences: []string{"dev.abcxyz.jvs"},
+			wantJustifications: []*jvspb.Justification{
```
sqin2019, line 194: nit: since there are multiple of them, can make this a common variable.

pkg/justification/processor_test.go
```diff
@@ -313,6 +364,12 @@ func TestCreateToken(t *testing.T) {
 			},
 			wantTTL:       1 * time.Hour,
 			wantAudiences: []string{DefaultAudience},
+			wantJustifications: []*jvspb.Justification{
```
sqin2019, line 367: nit: same here, can make this jira just a common variable.

sqin2019 (approved):

pkg/justification/processor_test.go
sqin2019: please ignore, data race

pkg/justification/processor_test.go
sqin2019: please ignore, data race

sqin2019 (approved):

PR #336, opened 2023-08-31
feat: Reset hint when click the reset button

### LINKED ISSUE #326: Reset doesn't reset the hint
Reset button only reset the category to "explanation" but it won't reset the hint.

qiaoweiz:
fix #326

### DIFF
--- a/assets/server/static/js/popup/main.js
+++ b/assets/server/static/js/popup/main.js
@@ -37,6 +37,14 @@ document.addEventListener("DOMContentLoaded", async function () {
 
   // Call the function when select new category.
   categorySelect.addEventListener("change", updatePlaceholder);
+
+  const resetButton = document.querySelector('#resetButton');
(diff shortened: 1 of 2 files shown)


assets/server/static/js/popup/main.js
```diff
@@ -37,6 +37,14 @@ document.addEventListener("DOMContentLoaded", async function () {
 
   // Call the function when select new category.
   categorySelect.addEventListener("change", updatePlaceholder);
+
+  const resetButton = document.querySelector('#resetButton');
+  resetButton.addEventListener("click", function () {
```
sethvargo, line 42: it would be more appropriate to bind to the form reset event: https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/reset_event

qiaoweiz: update in the latest version

Issue #327, opened 2023-08-11
Update current default category


qiaoweiz:
Current default category is explanation

PR #339, opened 2023-08-31
feat: handle default category option

qiaoweiz:
fix: #327 

- If there are multiple category options("Jira Issue Key", "Github" ...), then we don't want to show the default category option ("Explanation"). 
- If not, we can show the default category option.

### DIFF
--- a/apis/v0/plugin_test.go
+++ b/apis/v0/plugin_test.go
@@ -86,8 +86,8 @@ func TestGetUIDataInValidator(t *testing.T) {
 			name: "success",
 			req:  &GetUIDataRequest{},
 			wantResp: &UIData{
-				DisplayName: DefaultJustificationDisplayName,
-				Hint:        DefaultJustificationHint,
+				DisplayName: "Explanation",
+				Hint:        "A justification reason in free-form text.",
 			},
 		},
 	}
--- a/apis/v0/plugin_test.go
+++ b/apis/v0/plugin_test.go
@@ -86,8 +86,8 @@ func TestGetUIDataInValidator(t *testing.T) {
 			name: "success",
 			req:  &GetUIDataRequest{},
 			wantResp: &UIData{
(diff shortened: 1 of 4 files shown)


sethvargo (approved):

apis/v0/plugin_test.go
```diff
@@ -86,8 +86,8 @@ func TestGetUIDataInValidator(t *testing.T) {
 			name: "success",
 			req:  &GetUIDataRequest{},
 			wantResp: &UIData{
-				DisplayName: "Explanation",
-				Hint:        "A justification reason in free-form text.",
+				DisplayName: DefaultJustificationDisplayName,
+				Hint:        DefaultJustificationHint,
```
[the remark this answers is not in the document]
qiaoweiz: reverted changes in this file.

pkg/controller/controller.go
```diff
@@ -354,5 +355,22 @@ func catagoriesDisplayData(ctx context.Context, validators map[string]jvspb.Vali
 		}
 		displayData[k] = d
 	}
+
+	// In case there are additional category options available, we aim to hide the default option from users.
+	if len(displayData) > 1 {
+		delete(displayData, jvspb.DefaultJustificationCategory)
+	}
+
```
sqin2019, line 363: If the default category is removed here, what happen if in the popup user selects the default category?

sethvargo (approved):

pkg/controller/controller.go
qiaoweiz: If there are multiple category options("Jira Issue Key", "Github" ...), then we don't show the default category option ("Explanation") to users, users will not selects it. 
If there only one category option("Explanation"), the workflow when user select this option will be the same as before.

sqin2019: I see, thx!

sqin2019 (approved):

PR #340, opened 2023-09-06
fix: update ci module

sailorlqh:
Update CI module so the we can just use this module in `infra repo` to set up the necessary infra needed by CI tests.

### DIFF
--- a/terraform/ci/main.tf
+++ b/terraform/ci/main.tf
@@ -18,6 +18,10 @@ locals {
   github_repo_id  = 479173136 # abcxyz/jvs
 }
 
+resource "random_id" "default" {
+  byte_length = 2
+}
+
 resource "google_project_service" "services" {
   for_each = toset([
     "cloudkms.googleapis.com",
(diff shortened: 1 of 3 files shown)


terraform/ci/main.tf
```diff
@@ -29,30 +33,65 @@ resource "google_project_service" "services" {
...
 
-  role   = each.key
-  member = module.github_ci_infra.service_account_member
-}
+  location      = var.region
+  repository_id = var.registry_repository_id
+  description   = "Container registry for docker images."
+  format        = "DOCKER"
 
-module "github_ci_infra" {
```
sailorlqh, line 42: `github_ci_infra` creates wif and artifact registry, but now wif should be handled by [infra](https://github.com/abcxyz/infra/tree/main/terraform/workload-identity/github). So we remove this part, and create AR in CI, and use the `wif` created by infra is an input to bind with AR.

capri-xiyue: I think we have not moved jvs to the new strcuture handled by [infra](https://github.com/abcxyz/infra/tree/main/terraform/workload-identity/github). As you can see, there is no wif related repo to JVS https://github.com/abcxyz/infra/tree/main/terraform/workload-identity/github/abcxyz

sqin2019: +1, this is not safe to remove since https://github.com/abcxyz/infra/blob/main/terraform/deployment-resources/jvs-legacy/ci/main.tf#L6 does not pin the commit sha, so if anyone makes a change in infra, the wif will be removed, and we don't have one to replace.

sailorlqh: This PR is just show idea of the CI module looks like that's why it's in draft. I won't merge this until I have everything needed is set up in infra.

terraform/ci/main.tf
```diff
@@ -29,30 +33,55 @@ resource "google_project_service" "services" {
...
   kms_key_location = var.kms_key_location
 }
+
+module "jvs_services" {
+  source = "../modules/jvs-services"
+
+  project_id = var.project_id
+
+  region          = var.region
+  service_ingress = "all"
```
sailorlqh, line 74: This is set to all because we are not using GCLB in CI project. Authentication is still required to send request to the service, so this shouldn't be a problem.

terraform/ci/variables.tf
```diff
@@ -44,3 +44,30 @@ variable "ci_iam_roles" {
...
+  type        = map(string)
+  default     = {}
+}
+
+variable "registry_repository_id" {
+  description = "name for artifact registry."
+  type        = string
+}
+
+variable "ci_service_account_email" {
```
sailorlqh, line 70: We will use the WIF service account created by https://github.com/abcxyz/infra/pull/435 as the CI service account.

terraform/ci/variables.tf
```diff
@@ -44,3 +44,30 @@ variable "ci_iam_roles" {
...
+  type        = string
+  default     = "us-central1"
+}
+
+variable "jvs_container_image" {
+  description = "Container image for the jvsctl CLI and server entrypoints."
+  type        = string
+}
+
+variable "plugin_envvars" {
```
sqin2019, line 59: both "plugin_envvars" and "ci_service_account_email" variable should be removed. We wont test plugin in jvs ci.

sailorlqh: Moved.

sqin2019: oops, missed two more: variable "ci_iam_roles", "registry_repository_id"

sqin2019 (approved):

PR #344, opened 2023-09-08
fix: output service account id

sailorlqh:


lock14 (dismissed):

terraform/modules/common/outputs.tf
```diff
@@ -17,21 +17,41 @@ output "jvs_api_service_account_email" {
...
+}
+
 output "jvs_ui_service_account_email" {
   description = "JVS UI service account email."
   value       = google_service_account.ui_acc.email
 }
 
+output "jvs_ui_service_account_id" {
+  description = "JVS UI service account id."
+  value       = google_service_account.ui_acc.id
```
sqin2019, line 32: let's use name instead of id and rename every where. See https://registry.terraform.io/providers/hashicorp/google-beta/latest/docs/resources/google_service_account_iam#google_service_account_iam_member
Screenshot 2023-09-08 at 2 26 05 PM

sailorlqh: Updated

sqin2019 (approved):

Issue #348, opened 2023-09-12
Add clean up workflow to remove cloud run revisions.


sailorlqh:
The current CI workflow will create a revision for integration test, but it can't remove that revision because it's not allowed to remove the newest revision.

Therefore, a work around would be creating a cleanup workflow that is scheduled to run every day (or on other frequency) that automatically remove all the revisions that are older then a certain period e.g 12 hours.

sqin2019:
Do we need to cleanup gar as well, or images will expire themselves?

sailorlqh:
Yes, but that would be a different workflow other than this one to clean up revisions, because the image clean up only need to be run once to setup the auto clean up policy.  And we already have that in PMAP: https://github.com/abcxyz/pmap/blob/main/.github/workflows/update-ci-ar-cleanup-policy.yml

sqin2019:
We can configure that in infra when creating the gar right?

sailorlqh:
It's not supported by terraform yet. We can either manually run the `gcloud` command with our own account, or we can do as what PMAP does.

sqin2019:
@sailorlqh believe this is also done, closing, feel free to reopen

PR #349, opened 2023-09-12
fix: update ci workflow

sailorlqh:
Update ci workflow to deploy code to cloud run service, which will be used in future integration test.

Revision clean up will be in a different workflow and in a different PR. The reason for needing a cleanup workflow instead of just clean up in `CI` is stated in https://github.com/abcxyz/jvs/issues/348#issue-1893104350.

### DIFF
--- a/.github/workflows/ci-new.yml
+++ b/.github/workflows/ci-new.yml
@@ -1,0 +1,233 @@
+# Copyright 2023 The Authors (see AUTHORS file)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# We will remove ci.yml and rename this to ci.yml
+# after the new integration test implementation
+# has finished.
+name: 'ci-new'
+
+# This will also be updated once the new integration
+# test implementation has finished
+on:
+  # this pull_request here is just for me to
+  # test this workflow works before merging to main
+  # I will remove pull_request and only keep
+  # workflow_dispatch before merging this PR to main.
+  pull_request:
+    branches:
+      - 'main'
+  workflow_dispatch:
+
+env:
+  PROJECT_ID: 'jvs-i-e5'
+  WIF_PROVIDER: 'projects/851681819867/locations/global/workloadIdentityPools/github-automation/providers/jvs-ci-i'
+  WIF_SERVICE_ACCOUNT: 'github-automation-bot@gha-jvs-ci-i-31a91c.iam.gserviceaccount.com'
+  CONTAINER_REGISTRY: 'us-docker.pkg.dev/jvs-i-e5/ci-images'
+  DOCKER_TAG: '${{ github.sha }}'
+  IMAGE_NAME: 'jvsctl'
+
+  API_SERVICE_NAME: 'jvs-api-4527'
+  CERT_ROTATOR_SERVICE_NAME: 'jvs-cert-rotator-e04a'
+  PUBLIC_KEY_SERVICE_NAME: 'jvs-public-key-f22e'
+  UI_SERVICE_NAME: 'jvs-ui-8ab7'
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -68,10 +93,95 @@ jobs:
...
+  deployment:
+    runs-on: 'ubuntu-latest'
+    needs:
+      - 'build'
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+    strategy:
+      matrix:
+        services: ['jvs-api-4527', 'jvs-cert-rotator-e04a', 'jvs-public-key-f22e', 'jvs-ui-8ab7']
```
sailorlqh, line 156: Matrix doesn't support ${{ env.XXX }} so I have to hard code it here.
I think this is better than having four steps and each of them updates one services
```
- name: 'Update API service'
        run: gcloud run services update api_service
- name: 'Update cert rotator service'
        run: gcloud run services update cert_rotator
......
```

.github/workflows/ci.yml
```diff
@@ -68,10 +93,95 @@ jobs:
...
+        uses: 'actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744' # ratchet:actions/checkout@v3
+      - id: 'auth'
+        name: 'Authenticate to Google Cloud'
+        uses: 'google-github-actions/auth@35b0e87d162680511bf346c299f71c9c5c379033' # ratchet:google-github-actions/auth@v1
+        with:
+          workload_identity_provider: '${{ env.WIF_PROVIDER }}'
+          service_account: '${{ env.WIF_SERVICE_ACCOUNT }}'
+      - name: 'Setup gcloud'
+        uses: 'google-github-actions/setup-gcloud@e30db14379863a8c79331b04a9969f4c1e225e0b' # ratchet:google-github-actions/setup-gcloud@v1
+      - name: 'Update service'
```
sailorlqh, line 168: the revisions will be cleaned up in a different scheduled cleanup workflow. https://github.com/abcxyz/jvs/issues/348

.github/workflows/ci.yml
sqin2019: We cannot use matrix as we might need to add kms key as env var for these services, and they use different names for it.

.github/workflows/ci.yml
```diff
@@ -92,11 +202,5 @@ jobs:
       - name: 'Run integration tests'
         env:
           TEST_INTEGRATION: 'true'
-          TEST_JVS_KMS_KEY_RING: 'projects/${{ env.PROJECT_ID }}/locations/global/keyRings/ci-keyring'
         run: |-
-          go test \
```
sqin2019, line 97: I would prefer we keep the current test for now.

sailorlqh: The current test requires the KMS to have have a ci-keyring. We haven't created that yet.

https://github.com/abcxyz/jvs/blob/main/.github/workflows/ci.yml#L95

sqin2019: we can use the king ring in the legacy project, and add a comment that this test will be removed when the real integration test is added.

sailorlqh: That an option, but doing this would require us to grant the current wif SA roles like `roles/kms.admin` to the legacy CI test project. I think that not necessary as we are migrating away from the legacy project.

.github/workflows/ci-new.yml
```diff
@@ -0,0 +1,233 @@
...
+  CONTAINER_REGISTRY: 'us-docker.pkg.dev/jvs-i-e5/ci-images'
+  DOCKER_TAG: '${{ github.sha }}'
+  IMAGE_NAME: 'jvsctl'
+
+  API_SERVICE_NAME: 'jvs-api-4527'
+  CERT_ROTATOR_SERVICE_NAME: 'jvs-cert-rotator-e04a'
+  PUBLIC_KEY_SERVICE_NAME: 'jvs-public-key-f22e'
+  UI_SERVICE_NAME: 'jvs-ui-8ab7'
+  SERVICES_URL_PREFIX: 'https://'
+  SERVICES_URL_POSTFIX: '2nhpyabgtq-uc.a.run.app'
```
sqin2019, line 46: Remove?

sqin2019 (dismissed):

.github/workflows/ci-new.yml
sailorlqh: I removed the `SERVICES_URL_PREFIX`, but I think `SERVICES_URL_POSTFIX` is necessary as in integ test, we will need this to put together the whole service url with tag. Something like `tag---api-service-postfix:443`.

sqin2019 (approved):

PR #350, opened 2023-09-13
feat: add integ test using real service for API and Key service

sailorlqh:
Integration test for cert rotator service and ui services will be added in the follow up PRs.

part of the fix for #158


test/integ-new/main_new_test.go
```diff
@@ -0,0 +1,206 @@
...
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			ctx := context.Background()
+
+			var createCmd cli.TokenCreateCommand
+			_, stdout, _ := createCmd.Pipe()
+
+			server := fmt.Sprintf("%s---%s-%s:443", cfg.RevisionTagID, cfg.APIServiceName, cfg.ServiceURLPostfix)
```
sqin2019, line 137: for this and the public key endpoint - I wonder if we can pass them as single env vars in the github workflow, if not, better to create them in https://github.com/abcxyz/jvs/pull/350/files#diff-134ca5370506bf2f23ad76f5bfce577db50a4137f3c139339f06cd5497603879R36

test/integ-new/main_new_test.go
```diff
@@ -0,0 +1,206 @@
...
+----- Justifications -----
+breakglass    issues/12345
+
+----- Claims -----
+aud    [dev.abcxyz.jvs]
+exp    %s
+iat    %s
+iss    jvsctl
+sub
+`, now.Add(30*time.Minute).UTC().Format(timestampFormat), now.Format(timestampFormat)),
```
sqin2019, line 123: We can make ttl a global const and use it across all test cases, so that it is consistent.

test/integ-new/main_new_test.go
sailorlqh: Yes, this is doable, updated.

test/integ-new/main_new_test.go
sailorlqh: Done.

sqin2019 (dismissed):

test/integ-new/main_new_test.go
```diff
@@ -0,0 +1,196 @@
...
+)
+
+func TestMain(m *testing.M) {
+	os.Exit(func() int {
+		ctx := context.Background()
+
+		ignoreKeysMap = map[string]struct{}{
+			"nbf": {},
+			"jti": {},
+		}
```
yolocs, line 58: Why not directly define it in line 42?

test/integ-new/main_new_test.go
```diff
@@ -0,0 +1,196 @@
...
+			justification: "issues/12345",
+			isBreakglass:  false,
+			wantValidationResult: fmt.Sprintf(`
+----- Justifications -----
+explanation    issues/12345
+
+----- Claims -----
+aud    [dev.abcxyz.jvs]
+exp    %s
+iat    %s
```
yolocs, line 98: We should be able to use https://github.com/abcxyz/jvs/blob/main/pkg/cli/token_create.go#L60 to set a fixed "now"? With that, can we turn the check to be string comparison (just compare the output string rather than parsing them)?

yolocs (changes requested):

test/integ-new/main_new_test.go
sailorlqh: The reason for needing parsing is fields like `jti (jwt id)` that in the validation result is not known before creating the token, so we have to parse the field and ignore those fields.

Updated to use `--now` to make test more stable.

test/integ-new/main_new_test.go
sailorlqh: Done

test/integ-new/main_new_test.go
yolocs: I see. Would using `-format=json` flag make it easier to parse? https://github.com/abcxyz/jvs/blob/main/pkg/cli/token_validate.go#L86

sailorlqh: I have considered that, but I think this will require creating some strict and unmarshel the json, and then doing diff with ignoreField. I think this is more complicated then the current method. Or maybe I am missing something here?

yolocs: I think being strict is what we want here. Otherwise we'd be introducing flakiness in the test. If we change the output format a little bit, the test will break.

sailorlqh: I have changed to use `-format=json` and `cmp.Diff` to do the compare.

After talking to Suhong, `--now` can only apply to breakglass. And if we use `format=json`, the timestamp will be in `RFC3339`, so I updated the timestamp comparison to not compare the seconds to make tests less flaky.

test/integ-new/main_new_test.go
```diff
@@ -0,0 +1,212 @@
...
+
+			var validateCmd cli.TokenValidateCommand
+			_, stdout, _ = validateCmd.Pipe()
+
+			if err := validateCmd.Run(ctx, validateTokenArgs); err != nil {
+				t.Fatal(err)
+			}
+
+			gotTestValidationResult := &testValidationResult{}
+			if err := json.Unmarshal(stdout.Bytes(), gotTestValidationResult); err != nil {
```
sqin2019, line 172: What happens if the json string has some fields that are not found in the gotTestValidationResult struct? Maybe unmarshal it to a map is better.

test/integ-new/main_new_test.go
```diff
@@ -0,0 +1,212 @@
...
+
+	if want.Exp[:len(want.Exp)-3] != got.Exp[:len(got.Exp)-3] {
+		rErr = errors.Join(rErr, fmt.Errorf(fmt.Sprintf("got unexpected exp timestamp (-want, +got)\n -%s\n +%s\n", want.Exp, got.Exp)))
+	}
+
+	if want.Exp[:len(want.Iat)-3] != got.Exp[:len(got.Iat)-3] {
+		rErr = errors.Join(rErr, fmt.Errorf(fmt.Sprintf("got unexpected iat timestamp (-want, +got)\n -%s\n +%s\n", want.Iat, got.Iat)))
+	}
+
+	if want.Exp[:len(want.Iss)-3] != got.Exp[:len(got.Iss)-3] {
```
sqin2019, line 208: Is this a typo? Iss is issuer instead of timestamp. We probably should not ignore compare this field as well.

sailorlqh: Fixed.

test/integ-new/main_new_test.go
sqin2019: Or should we use jwt.Token instead?

sqin2019: Did a quick check, it turns out we are using https://github.com/abcxyz/jvs/blob/9a2eff640dced585729be84a7a53f71869bd61d4/pkg/formatter/formatter.go#L36 for the json formatter.

yolocs: I thought more about this. For context, this formatted structure _is_ defined here: https://github.com/abcxyz/jvs/blob/main/pkg/formatter/formatter.go#L36 except that it's not public.
However, I don't think we should run the test based on the output. The output format is simply a representation of the input data and it can be fully covered in unit test.
What we really care here is whether the running service can give us token and validate token. I recommend updating the test:

1. Run first command to create a token. Ensure success.
2. Run second command to validate the token. Ensure success. Don't check the output.
3. Parse the token to ensure required claims are there.

The only difference is the 3rd step. Instead of checking the output which is up to change, we check the token directly.

yolocs (changes requested):

test/integ-new/main_new_test.go
sqin2019: Ok, so we are probably fine here.

test/integ-new/main_new_test.go
```diff
@@ -0,0 +1,209 @@
...
+			var createCmd cli.TokenCreateCommand
+			_, stdout, _ := createCmd.Pipe()
+
+			createTokenArgs := []string{"-server", cfg.APIServer, "-justification", tc.justification, "-ttl", testTTLString, "--auth-token", cfg.IDToken}
+			if tc.isBreakglass {
+				createTokenArgs = append(createTokenArgs, "-breakglass")
+			}
+
+			if err := createCmd.Run(ctx, createTokenArgs); err != nil {
+				t.Fatal(err)
```
sqin2019, line 157: nit: add more error details like "failed to create token"

test/integ-new/main_new_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+			token := stdout.String()
+
+			validateTokenArgs := []string{"-token", token, "-jwks-endpoint", cfg.JWTEndpoint, "--format", "json"}
+
+			var validateCmd cli.TokenValidateCommand
+			_, stdout, _ = validateCmd.Pipe()
+
+			if err := validateCmd.Run(ctx, validateTokenArgs); err != nil {
+				t.Fatal(err)
```
sqin2019, line 168: nit: add more error details like "failed to validate token"

test/integ-new/main_new_test.go
sqin2019: To clarify, it is doing the three things in a single test func, or do you mean to split the test func into 2, one for create token and one for validate token?

sqin2019 (dismissed):

test/integ-new/main_new_test.go
```diff
@@ -154,56 +149,58 @@ func TestAPIAndPublicKeyService(t *testing.T) {
...
+			_, _, _ = validateCmd.Pipe()
+
+			if err := validateCmd.Run(ctx, validateTokenArgs); err != nil {
+				t.Fatalf("jvs service failed to validate token: %v", err)
 			}
 
-			if err := testDiffClaimTimestamp(t, tc.wantTestValidationResult.Claims, gotTestValidationResult.Claims); err != nil {
-				t.Errorf("Claims got unexpected diff: %v", err)
+			if diff := cmp.Diff(testParseTokenMap(t, tc.wantTokenMap), testParseTokenMap(t, parsedTokenMap)); diff != "" {
+				t.Errorf("token got unexpacted diff (-want, +got):\n%s", diff)
```
yolocs, line 177:
```suggestion
				t.Errorf("token got unexpected diff (-want, +got):\n%s", diff)
```

test/integ-new/main_new_test.go
```diff
@@ -0,0 +1,206 @@
...
+			}
+		})
+	}
+}
+
+// testParseToken parses the tokenMap by overwriting the value for ignoreKeys
+// to empty and parsing the timestamp to the format without seconds to make
+// test on timestamps less flaky.
+func testParseTokenMap(tb testing.TB, m map[string]any) map[string]any {
```
yolocs, line 186: nit: something like `testNormalizeTokenMap` would make more sense.

yolocs (approved):

sqin2019 (approved):

PR #355, opened 2023-09-25
fix: change /healthz to /health to fix cloud run known issue

sailorlqh:
Url ending with `z` could cause trouble. 
See: https://cloud.google.com/run/docs/issues#ah

We can also use `health` as the endpoint for cloud run  HTTP liveness probe in the future if we want to.

### DIFF
--- a/pkg/cli/public_key_server.go
+++ b/pkg/cli/public_key_server.go
@@ -119,7 +119,10 @@ func (c *PublicKeyServerCommand) RunUnstarted(ctx context.Context, args []string
 	keyServer := jvscrypto.NewKeyServer(ctx, kmsClient, c.cfg, h)
 
 	mux := http.NewServeMux()
-	// This is the healthz checkpoint.
+
+	// Use /health instead of /healthz to avoid cloud run reserved path
+	//
+	// See: https://cloud.google.com/run/docs/issues#ah
 	mux.Handle("/health", healthcheck.HandleHTTPHealthCheck())
 	mux.Handle("/.well-known/jwks", keyServer)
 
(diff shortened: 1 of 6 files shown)


pkg/cli/public_key_server.go
```diff
@@ -119,7 +119,11 @@ func (c *PublicKeyServerCommand) RunUnstarted(ctx context.Context, args []string
 	keyServer := jvscrypto.NewKeyServer(ctx, kmsClient, c.cfg, h)
 
 	mux := http.NewServeMux()
-	mux.Handle("/healthz", healthcheck.HandleHTTPHealthCheck())
+	// This is the healthz checkpoint, but in cloud run url ending with z
+	// could be reserved. This is to fix that.
+	//
+	// See: https://cloud.google.com/run/docs/issues#ah
+	mux.Handle("/healthz-check", healthcheck.HandleHTTPHealthCheck())
```
sethvargo, line 126: Why not `/health` or `/.operations/health`?

pkg/cli/public_key_server.go
```diff
@@ -119,7 +119,8 @@ func (c *PublicKeyServerCommand) RunUnstarted(ctx context.Context, args []string
 	keyServer := jvscrypto.NewKeyServer(ctx, kmsClient, c.cfg, h)
 
 	mux := http.NewServeMux()
-	mux.Handle("/healthz", healthcheck.HandleHTTPHealthCheck())
+	// This is the healthz checkpoint.
```
sqin2019, line 122: nit: seems redundant. same to the other comments. maybe just add the reference link and mention to avoid cloud run reserved paths.

sqin2019 (dismissed):

pkg/cli/public_key_server.go
sailorlqh: Changed to `health`, previous I had some misunderstanding, I thought `health` is reserved for liveness probe.

Thank you!

sqin2019 (approved):

yolocs (approved):
Can we check if we need to update GCLB health check path?

PR #357, opened 2023-09-27
feat: add integ test for kms service

sailorlqh:


.github/workflows/ci-new.yml
```diff
@@ -160,13 +161,23 @@ jobs:
           service_account: '${{ env.WIF_SERVICE_ACCOUNT }}'
       - name: 'Setup gcloud'
         uses: 'google-github-actions/setup-gcloud@e30db14379863a8c79331b04a9969f4c1e225e0b' # ratchet:google-github-actions/setup-gcloud@v1
+      - name: 'create_kms_key'
```
sqin2019, line 164: Do we need to cleanup the keys?

sailorlqh: I am thinking we will have a dedicated workflow to auto cleanup the cloud run revisions, we should cleanup the keys in the same place as they are both cleanup work.

sqin2019: Did a quick search, a key cannot be deleted, and we are charged for the number of active key versions, we should disable the key version in each run for cleanup purpose. It seems easier to do it here in this workflow. Also have you check is there a limit on how many keys we can have in a key ring, or if any issues with creating a lot of keys in a single key ring?

sailorlqh: Oh I will take a look on this and update the PR if necessary

sailorlqh: Updated the code, now we destroy keys in the CI workflow.
See: https://screenshot.googleplex.com/3AaQm3vUjHXUqLK

.github/workflows/ci-new.yml
```diff
@@ -248,3 +273,27 @@ jobs:
             -shuffle=on \
             -timeout='10m' \
             ./test/integ-new/...
+
+  cleanup:
```
sqin2019, line 277: What if integration failed in the beginning and the key is not yet created. I would suggest just add a step in integration job at the end and the condition would be `if: '${{ always() && steps.create_kms_key.outcome == ''success'' }}'`, note you need to set a step id `id: create_kms_key` in your create_kms_key step.

.github/workflows/ci-new.yml
```diff
@@ -273,24 +274,8 @@ jobs:
...
-      - name: 'Checkout'
-        uses: 'actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b' # ratchet:actions/checkout@v3
-      - id: 'auth'
-        name: 'Authenticate to Google Cloud'
-        uses: 'google-github-actions/auth@35b0e87d162680511bf346c299f71c9c5c379033' # ratchet:google-github-actions/auth@v1
-        with:
-          workload_identity_provider: '${{ env.WIF_PROVIDER }}'
-          service_account: '${{ env.WIF_SERVICE_ACCOUNT }}'
-      - id: 'delete_kms_key'
+      - name: "clean up kms key"
```
sqin2019, line 277: Cloud you also add comment that we cannot delete key, therefore we are destroying key versions so that we are not billed, and add the reference links?

.github/workflows/ci-new.yml
```diff
@@ -248,3 +274,14 @@ jobs:
             -shuffle=on \
             -timeout='10m' \
             ./test/integ-new/...
+      # Keys can not be deleted, we can only destroy them.
+      # We are not charged by keys that are destried.
+      # See: https://cloud.google.com/kms/docs/faq#cannot_delete.
+      - name: 'clean up kms key'
+        if: '${{ always()  && needs.deployment.steps.create_kms_key.outcome == ''success'' }}'
```
sqin2019, line 281: if: '${{ always() && steps.create_kms_key.outcome == ''success'' }}'

Also, -> "Keys cannot be deleted, we can destroy key versions to avoid billing"

sqin2019 (dismissed):

sqin2019 (approved):

test/integ-new/main_new_test.go
```diff
@@ -220,6 +220,63 @@ func TestUIServiceHealthCheck(t *testing.T) {
...
+		path           string
+		wantStatusCode int
+	}{
+		{
+			name:           "cert_rotator_health_check",
+			path:           "/health",
+			wantStatusCode: http.StatusOK,
+		},
+		{
+			name:           "cert_rotator_key_rotation",
```
yolocs, line 237: It doesn't seem we verify the key was actually rotated? Shouldn't we do that?

sqin2019: IIRC, the key won't actually rotate since the current version not expired yet. We can confirm the primary key is still the first version if you think it is necessary.

sailorlqh: +1

yolocs: So are you saying we're going to lose the test coverage in https://github.com/abcxyz/jvs/blob/main/test/integ/main_test.go#L326?

sqin2019: right, should these be covered in unit test instead?

yolocs: How to cover them in unit test? We won't connect to KMS in unit test

sqin2019: True, then we need to add some more test cases.

sailorlqh: Create issue https://github.com/abcxyz/jvs/issues/360.  Will do this as a follow up.

yolocs: Discussed offline. Will follow up in a separate PR

yolocs (approved):

PR #358, opened 2023-09-28
feat: add clean up workflow to remove old cloud run revisions

sailorlqh:


.github/workflows/cleanup.yml
```diff
@@ -0,0 +1,154 @@
...
+            --filter="metadata.creationTimestamp < '-p5h'" | sed 1d)
+
+          IFS=$'\n'; for NAME in $REVISIONS; do
+            echo "Deleting ${NAME}..."
+            gcloud run revisions delete ${NAME} --project="$INTEGRATION_PROJECT_ID" --region="$INTEGRATION_REGION" --quiet --async
+          done
+
+  # cleanup_cert_rotator_service_cloudrun_revisions deletes all Cert Rotator service Cloud Run revisions for the given service
+  # that are more than 5 hours old
+  cleanup_cert_rotator_service_cloudrun_revisions:
```
capri-xiyue, line 71: Can we use https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs to avoid such duplications? Looks like the only difference between different jobs are the `--service`?

sqin2019: +1

sailorlqh: Yes, that's doable, my idea was GitHub metrics aggregator also divided this into different steps instead of using a matrix, it might be better to keep the consistency?

capri-xiyue: Do you know the reason of GitHub metrics aggregator not using matrix. If there is no specific reason, probably it's better to change the GitHub metrics aggregator to use matrix as it makes the code cleaner by avoiding the duplication?

sailorlqh: Confirmed with Mike that there isn’t a specific reason for not using matrix. I will update this PR to use matrix later.

sailorlqh: Updated to use matrix

yolocs (approved):

sqin2019 (approved):

capri-xiyue (approved):

PR #367, opened 2023-10-09
chore: use different keyring for differney ci run

sailorlqh:


sqin2019 (dismissed):

.github/workflows/ci-new.yml
```diff
@@ -36,7 +36,7 @@ env:
   PUBLIC_KEY_SERVICE_NAME: 'jvs-public-key-f22e'
   UI_SERVICE_NAME: 'jvs-ui-8ab7'
   SERVICES_URL_POSTFIX: '2nhpyabgtq-uc.a.run.app'
-  KMS_KEYRING_ID: 'jvs-keyring-0666'
+  KMS_KEYRING_ID: 'jvs-keyring-ci-${{ github.run_id }}-${{ github.run_number }}'
```
verbanicm, line 39: Just curious, if this workflow fails and you click re-run, do you expect this number to be different? If yes, then you should use `github.run_attempt` instead of `run_number`.

Otherwise, LGTM

.github/workflows/ci-new.yml
```diff
@@ -165,6 +165,12 @@ jobs:
           service_account: '${{ env.WIF_SERVICE_ACCOUNT }}'
       - name: 'Setup gcloud'
         uses: 'google-github-actions/setup-gcloud@e30db14379863a8c79331b04a9969f4c1e225e0b' # ratchet:google-github-actions/setup-gcloud@v1
+      - id: 'create_kms_keyring'
+        name: 'Create kms keyring'
+        run: |-
+          gcloud kms keyrings create ${{ env.KMS_KEYRING_ID }} \
+            --project ${{ env.PROJECT_ID }} \
+            --location global
```
sethvargo, line 173: Hey @sailorlqh - can you share more info about why we need to create a new keyring on each CI run? It seems like we should be able to create a cryptokey and N cryptokeyversions for each CI run instead.

sailorlqh: The current way is create a new ring in the same keyring. The reason for this is discussed in the infra repo: https://github.com/abcxyz/infra/pull/552#discussion_r1352873887

sailorlqh:
Close this PR as we agreed we will use 1 keyring with multiple keys in it.

PR #372, opened 2023-10-12
feat: add edge cases for cert rotator integ test

sailorlqh:


test/integ-new/main_new_test.go
```diff
@@ -303,6 +304,38 @@ func TestCertRotatorKeyRotation(t *testing.T) {
 				3: kmspb.CryptoKeyVersion_ENABLED,
 			})
 	})
+
+	t.Run("invalid_primary", func(t *testing.T) {
+		// Set primary to a version that doesn't exist
+		if err := jvscrypto.SetPrimary(ctx, kmsClient, keyResouceName, keyResouceName+"/cryptoKeyVersions/99"); err != nil {
```
sailorlqh, line 310: if we don't set a label or setup a invalid label in ci workflow (https://github.com/abcxyz/jvs/blob/main/.github/workflows/ci-new.yml#L183), the public key's none-breakglass test case will fail. 

Therefore, for these edge cases, I think it's better set label after the normal cases have finished, and then run the edge cases. 

In theory this way the rest of the integ test cases are already finished when cert rotator's test reaches this step. If we have concerns, we can consider sequently running public key service's integ test and then run cert rotator's.

sethvargo, line 310: Is this actually possible in the upstream API? I don't think KMS will let you set the primary version to a non-existent thing.

test/integ-new/main_new_test.go
```diff
@@ -401,3 +434,44 @@ func testValidateKeyVersionState(ctx context.Context, tb testing.TB, kmsClient *
...
+	key.Labels = labels
+
+	var mT *kmspb.CryptoKey
+	mask, err = fieldmaskpb.New(mT, "labels")
+	if err != nil {
+		tb.Fatalf("unable to create field mask: %s", err)
+	}
+	if _, err := kmsClient.UpdateCryptoKey(ctx, &kmspb.UpdateCryptoKeyRequest{CryptoKey: key, UpdateMask: mask}); err != nil {
+		tb.Fatalf("unable to set labels: %s", err)
+	}
```
sethvargo, line 476: You can set the labels and state in the same RPC. Just add both fields and use both in the update mask:

```go
        req := &kmspb.UpdateCryptoKeyVersionRequest{
                CryptoKeyVersion: &kmspb.CryptoKeyVersion{
                        Name:  name,
                        State: kmspb.CryptoKeyVersion_DISABLED,
                        Labels: labels,
                },
                UpdateMask: &fieldmask.FieldMask{
                        Paths: []string{"state", "labels"},
                },
        }
```

test/integ-new/main_new_test.go
sailorlqh: I think the primary is just a label, we can set it like this: https://github.com/abcxyz/jvs/blob/main/.github/workflows/ci-new.yml#L183

test/integ-new/main_new_test.go
```diff
@@ -303,6 +304,38 @@ func TestCertRotatorKeyRotation(t *testing.T) {
 				3: kmspb.CryptoKeyVersion_ENABLED,
 			})
 	})
+
+	t.Run("invalid_primary", func(t *testing.T) {
+		// Set primary to a version that doesn't exist
```
yolocs, line 309: My memory is failing me, which means it's something we should document better - What's the purpose of setting primary to a non-existing version? I see your comment below but I can't understand it.

sqin2019: Maybe a better edge case is when primary label is missing.

test/integ-new/main_new_test.go
sethvargo: Ah I see - we're not setting the "primary" field because it's an asymmetric key?

test/integ-new/main_new_test.go
sailorlqh: updated to remove the label

test/integ-new/main_new_test.go
```diff
@@ -303,6 +305,43 @@ func TestCertRotatorKeyRotation(t *testing.T) {
...
+		// if err := jvscrypto.SetPrimary(ctx, kmsClient, keyResouceName, keyResouceName+"/cryptoKeyVersions/99"); err != nil {
+		// 	t.Fatalf("unable to set primary: %s", err)
+		// }
```
yolocs, line 313: Remove?

test/integ-new/main_new_test.go
sailorlqh: I think `state` is for a key version, and `labels` is for a key. [CryptoKey]https://pkg.go.dev/cloud.google.com/go/kms@v1.15.3/apiv1/kmspb#CryptoKey doesn't have a `state` field and [CryptoKeyVersion](https://pkg.go.dev/cloud.google.com/go/kms@v1.15.3/apiv1/kmspb#CryptoKeyVersion) doesn't have a `labels` field.

test/integ-new/main_new_test.go
sailorlqh: Done.

test/integ-new/main_new_test.go
sethvargo: Ah you're right, sorry. What labels are we removing here - just `primary`? This is deleting all labels. Also, it might be better to remove the lable and _then_ delete the key. Otherwise you have a race condition where the label is pointing to a disabled key.

sailorlqh: Thanks for pointing out the race condition, I have update the code.

test/integ-new/main_new_test.go
sailorlqh: Yes

sethvargo (dismissed):

test/integ-new/main_new_test.go
```diff
@@ -303,6 +305,38 @@ func TestCertRotatorKeyRotation(t *testing.T) {
...
+			map[int]kmspb.CryptoKeyVersion_CryptoKeyVersionState{
+				1: kmspb.CryptoKeyVersion_DESTROY_SCHEDULED,
+				2: kmspb.CryptoKeyVersion_ENABLED,
+				3: kmspb.CryptoKeyVersion_ENABLED,
+			})
+	})
+
+	t.Run("emergent_disable", func(t *testing.T) {
+		// Emergently disable our primary.
+		testEmergentDisable(ctx, t, kmsClient, keyResouceName, keyResouceName+"/cryptoKeyVersions/2")
```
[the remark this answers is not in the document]
sailorlqh: Done.

test/integ-new/main_new_test.go
```diff
@@ -401,3 +435,65 @@ func testValidateKeyVersionState(ctx context.Context, tb testing.TB, kmsClient *
 		tb.Errorf("primary was set to version %d, but expected %d", primaryNumber, expectedPrimary)
 	}
 }
+
+func testRemoveLabelPrimary(ctx context.Context, tb testing.TB, kmsClient *kms.KeyManagementClient, keyName string) error {
+	tb.Helper()
+	req := &kmspb.UpdateCryptoKeyRequest{
```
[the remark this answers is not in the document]
sailorlqh: I thought the purpose of the `invalid_primary` is remove the primary label, to test if the primary key is missing, the service will assign key 3 as the primary key. As you mentioned here: https://github.com/abcxyz/jvs/pull/372#discussion_r1357441987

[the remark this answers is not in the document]
sqin2019: maybe a refactor to use this func in testEmergentDisable?

sqin2019 (dismissed):

sqin2019 (approved):

PR #375, opened 2023-10-18
feat: add key clean up workflow

sailorlqh:
The purpose of this is to cleanup the keys created in the ci workflow that were not destroyed.

I have thought about including this in the `cleanup` workflow as a cron job, but currently there's no way to check if a key has active keyversions. So if we are going to use the cron job, we need to go through all keys and disable key version in those keys. It could take a lot of time in the future if we have ran too many times for the ci workflow.

I have enabled the option to destroy all key versions for all keys in the keyring when the input `key_id` is empty.


.github/workflows/kms-key-cleanup.yml
```diff
@@ -0,0 +1,90 @@
...
+
+          IFS=$'\n'; for VERSION in $KEY_VERSIONS; do
+            echo "Deleting ${VERSION}"
+            gcloud kms keys versions destroy ${VERSION} --quiet
+          done
+      - name: 'Destroy all key version in the keyring'
+        if: '${{ inputs.key_id == '''' }}'
+        shell: 'bash'
+        run: |-
+          KEY=$(gcloud kms keys list \
```
sqin2019, line 72: I wonder if we can add a time filter to this, otherwise, we don't need to query all keys everytime.

.github/workflows/kms-key-cleanup.yml
```diff
@@ -0,0 +1,90 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+name: 'cleanup'
+
+# When key_id is empty, we cleanup all key's keyversion in the keyring
+# and when key_id is provided, we only cleanup keyversion in that specific key.
+on:
```
sqin2019, line 20: Can we also add a schedule so that it is cleanup automatically?

.github/workflows/cleanup.yml
```diff
@@ -68,3 +68,8 @@ jobs:
             echo "Deleting ${NAME}..."
             gcloud run revisions delete ${NAME} --project="$INTEGRATION_PROJECT_ID" --region="$INTEGRATION_REGION" --quiet --async
           done
+  key_cleanup:
+    uses: './.github/workflows/kms-key-cleanup.yml'
```
sqin2019, line 72: What about just use the workflow instead of call a reusable work flow.

sailorlqh: Done

.github/workflows/cleanup.yml
```diff
@@ -68,3 +74,46 @@ jobs:
...
+      - name: 'create time span env var'
+        if: '${{ github.event_name == ''workflow_dispatch''}}'
+        shell: 'bash'
+        run: |-
+          echo "TIME_SPAN=${{ inputs.time_span }}" >> $GITHUB_ENV
+      - name: 'create time span env var'
+        if: '${{ github.event_name != ''workflow_dispatch''}}'
+        shell: 'bash'
+        run: |-
+          echo "TIME_SPAN=1d" >> $GITHUB_ENV
```
sqin2019, line 95: Nit: can we combine it with below, and use it in `Destroy all key version in the keyring` directly?
```
if [ $A = "B" ]; then
>   export a=b
> else
>   export a=b
> fi
```

sqin2019 (approved):

sqin2019 (approved):

PR #379, opened 2023-11-13
fix: add plugin_secret_envvars so CR service can get secret directly

sailorlqh:
Add `plugin_secret_envvars` variables so we can use the [secret_envvars](https://github.com/abcxyz/terraform-modules/blob/main/modules/cloud_run/main.tf#L121C11-L133) from `abcxyz/terraform-module/cloud_run`, and get secret from secret manager directly.

### DIFF
--- a/terraform/e2e/main.tf
+++ b/terraform/e2e/main.tf
@@ -71,9 +71,10 @@ module "jvs_services" {
 
   jvs_container_image = var.jvs_container_image
 
-  kms_keyring_id = module.jvs_common.kms_keyring_id
-  kms_key_name   = "signing-${random_id.default.hex}"
-  plugin_envvars = var.plugin_envvars
+  kms_keyring_id        = module.jvs_common.kms_keyring_id
+  kms_key_name          = "signing-${random_id.default.hex}"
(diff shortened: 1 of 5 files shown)


sqin2019 (approved):

sethvargo (approved):
You'll need to make sure the CR service account has access to the underlying secrets or else it will fail to boot.

sailorlqh:
> You'll need to make sure the CR service account has access to the underlying secrets or else it will fail to boot.

Yes, the plugin already has that covered: https://github.com/abcxyz/jvs-plugin-github/blob/main/terraform/modules/secret_manager/main.tf#L63

sqin2019:
> > You'll need to make sure the CR service account has access to the underlying secrets or else it will fail to boot.
> 
> Yes, the plugin already has that covered: https://github.com/abcxyz/jvs-plugin-github/blob/main/terraform/modules/secret_manager/main.tf#L63

We should work on an example infra setup in jvs-plugin-github, we may see circular dependencies of the jvs e2e module and the secret_manager module (jvs e2e need the secret version and name, secret manager need the CR sa created in e2e).

PR #381, opened 2023-11-15
feat: add plugin_envvar and plugin_secret_envvar for ci

sailorlqh:


sqin2019:
why this is required? we don't test plugin in jvs ci.

sailorlqh:
Close this PR, discussed offline, found a better way to achieve my purpose.

PR #382, opened 2023-11-28
chore: use gci for import ordering

sailorlqh:
Golang-CI change is introduced to abcxyz/pkg by: https://github.com/abcxyz/pkg/pull/247


yolocs (approved):
Thanks!

PR #388, opened 2024-02-13
Deprecate client-lib

sethvargo:
This removes the client-lib package and embeds the functionality directly into the generated protos. Originally, there was a goal of releasing the client independently from the server, but that proved to be complex and unnecessary. We undid that change, but kept the client separate.

This moves the logic into the generated protos where they belong, but maintains compatibility by aliasing the types and functions inside client-lib.

### DIFF
--- a/apis/v0/client.go
+++ b/apis/v0/client.go
@@ -1,0 +1,89 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


apis/v0/client.go
```diff
@@ -0,0 +1,89 @@
+// Copyright 2022 Google LLC
```
sethvargo, line 1: These are moved, but enough of the content changed that git isn't handling it.

yolocs (approved):

PR #409, opened 2024-03-22
Fix protogetter lint issues

pdewilde:


pdewilde:
Seems like GH is limited to # of lint errors it can  show at a time, will fix them all by running locally

sethvargo (approved):

PR #415, opened 2024-09-30
feat: migration of minty configuration

bradegler:
Migration of: https://github.com/abcxyz/abcxyz-services/blob/main/github-token-minter/deployments/configs/abcxyz/jvs.yaml

### DIFF
--- a/.github/minty.yaml
+++ b/.github/minty.yaml
@@ -1,0 +1,32 @@
+version: 'minty.abcxyz.dev/v2'
+
+rule:
+  if: |-
+    assertion.iss == 'https://token.actions.githubusercontent.com' &&
+    assertion.repository_owner_id == '93787867' &&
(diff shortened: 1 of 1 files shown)


.github/minty.yaml
```diff
@@ -0,0 +1,32 @@
+version: 'minty.abcxyz.dev/v2'
+
+rule:
+  if: |-
+    assertion.iss == 'https://token.actions.githubusercontent.com' &&
+    assertion.repository_owner_id == '93787867' &&
+    assertion.repository_id == '479173136' &&
+    assertion.ref == 'refs/heads/main' &&
+    assertion.workflow_ref == assertion.job_workflow_ref
```
verbanicm, line 9: thought: what happens when we have a workflow in the future who cant use these attributes? that could be a large refactor. Would it be a good idea to create a guide on what are the best top level attributes, etc.

PR #430, opened 2025-04-21
fix: lint issues

dcreey:
Temporarily adds the lint.yml from abcxyz/actions, then fixes lint problems, then removes the lint.yml. This repo will be added to the org ruleset so it gets linted after this merges.

Tested expected java format lint fail: https://github.com/abcxyz/jvs/actions/runs/14621743965/job/41023298247?pr=430


verbanicm (approved):

.github/workflows/release.yml
```diff
@@ -133,10 +133,7 @@ jobs:
     steps:
       - uses: 'actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683' # ratchet:actions/checkout@v4
 
-      - uses: 'actions/setup-java@7a6d8a8234af8eb26422e24e3006232cccaa061b' # ratchet:actions/setup-java@v4
-        with:
-          java-version: '11'
-          distribution: 'adopt'
+      - uses: 'abcxyz/actions/.github/actions/setup-java-strict@4fc7cc98f78e76307d3f1ee4367a4db03aef429c' # ratchet:exclude
```
sethvargo: This is excluded but pinned. That means it will _never_ be updated.

sethvargo (approved):

